PR #85, opened 2021-05-10
Fix waiting for service in clustered environment

worekleszczy:
Switch subscribing to service from message adapter to child actor - all context message adapters internally use single FuctionRef and this made cluster receptionist once watch over this actor twice with two different messages types which yields to runtime exception.


extension/src/main/scala/io/scalac/mesmer/extension/AkkaStreamMonitoring.scala
```diff
@@ -290,6 +290,8 @@ final class AkkaStreamMonitoring(
 
   private def waitForStart(): Behavior[Command] = {
     setTimeout()
+    log.debug("Waiting for start")
```
jczuchnowski, line 293: maybe we should be more descriptive about what we are waiting for in the logs

extension/src/main/scala/io/scalac/mesmer/extension/util/GenericBehaviors.scala
```diff
@@ -25,37 +26,32 @@ object GenericBehaviors {
...
+
         import context._
 
+        setLoggerName(classTag[I].runtimeClass)
+
         def start(): Behavior[Any] = {
 
-          val adapter = context.messageAdapter[Listing](identity)
+          /*
+            We use child actors instead of message adapters to avoid issues with clustered receptionist
```
jczuchnowski, line 37: We need a better description of the clustered receptionist problem.

extension/src/main/scala/io/scalac/mesmer/extension/util/GenericBehaviors.scala
```diff
@@ -8,6 +8,7 @@ import akka.actor.typed.receptionist.ServiceKey
 import akka.actor.typed.scaladsl.Behaviors
 
 import scala.reflect.ClassTag
+import scala.reflect.classTag
```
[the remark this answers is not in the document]
worekleszczy: I wonder why scalafix didn't convert this to import `scala.reflect.{ClassTag, classTag}` 🤔

extension/src/main/scala/io/scalac/mesmer/extension/util/GenericBehaviors.scala
worekleszczy: I'll come up with something

worekleszczy: I left a comment on why messageAdapters can't be use, but this is not specific to this fix but to working with receptionist. I guess there is no point of explaining reason there in depth as this solution is still valid use of actors.

jczuchnowski (approved):

PR #86, opened 2021-05-10
Export actor tree

worekleszczy:
Adds akka management plugin for exporting actor tree in JSON format


management/src/main/scala/io/scalac/mesmer/management/service/ActorTreeRoutesProvider.scala
```diff
@@ -0,0 +1,153 @@
...
+import com.typesafe.config.Config
+import io.scalac.mesmer.core.model.Tag
+import io.scalac.mesmer.extension.config.Configuration
+import io.scalac.mesmer.extension.service.ActorTreeService.Command.GetActors
+import io.scalac.mesmer.extension.service.{ actorTreeServiceKey, ActorTreeService, NonEmptyTree }
+import io.scalac.mesmer.extension.util.GenericBehaviors
+import io.scalac.mesmer.management.service.ActorInfoService.{ actorInfo, ActorInfo }
+import org.slf4j.LoggerFactory
+import zio.Chunk
+import zio.json._
```
jczuchnowski, line 24: It looks like Akka Management depends on akka-http-spray-json, so we might not need any other dependency at all https://repo1.maven.org/maven2/com/lightbend/akka/management/akka-management_2.13/1.1.0/akka-management_2.13-1.1.0.pom

lgajowy:
@jczuchnowski do we still need this?

PR #328, opened 2022-02-24
Opentelemetry configuration

### LINKED ISSUE #301: New configuration approach

worekleszczy:
Closes #301 
___

## Summary:

Tries to replace our current shared configuration with OT one. This doesn't cover all of the features e.g. setting configuration after class transformations are registered, but we think that this is not a critical issue for us.

### DIFF
--- a/agent/src/main/scala/io/scalac/mesmer/agent/akka/actor/AkkaActorAgent.scala
+++ b/agent/src/main/scala/io/scalac/mesmer/agent/akka/actor/AkkaActorAgent.scala
@@ -1,26 +1,20 @@
 package io.scalac.mesmer.agent.akka.actor
 
-import net.bytebuddy.asm.Advice
-import net.bytebuddy.dynamic.TargetType
-import net.bytebuddy.implementation.FixedValue
-import net.bytebuddy.implementation.MethodCall
-import net.bytebuddy.implementation.MethodDelegation
-import net.bytebuddy.implementation.SuperMethodCall
-
-import io.scalac.mesmer.agent.Agent
-import io.scalac.mesmer.agent.AgentInstrumentation
+import io.scalac.mesmer.agent.{ Agent, AgentInstrumentation }
 import io.scalac.mesmer.agent.akka.actor.impl._
 import io.scalac.mesmer.agent.util.i13n._
-import io.scalac.mesmer.core.actor.ActorCellDecorator
-import io.scalac.mesmer.core.actor.ActorCellMetrics
+import io.scalac.mesmer.core.actor.{ ActorCellDecorator, ActorCellMetrics }
 import io.scalac.mesmer.core.akka.version26x
 import io.scalac.mesmer.core.model._
 import io.scalac.mesmer.core.module.AkkaActorModule
 import io.scalac.mesmer.core.util.Timestamp
+import net.bytebuddy.asm.Advice
+import net.bytebuddy.dynamic.TargetType
+import net.bytebuddy.implementation.{ FixedValue, MethodCall, MethodDelegation, SuperMethodCall }
 
 object AkkaActorAgent
     extends InstrumentModuleFactory(AkkaActorModule)
-    with AkkaActorModule.All[AkkaActorModule.Jars[Version] => Option[Agent]]
+    with AkkaActorModule.All[AkkaActorModule.Jars[Version] => Agent]
     with AkkaMailboxInstrumentations {
 
   /**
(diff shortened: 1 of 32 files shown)


core/src/main/scala/io/scalac/mesmer/core/module/Module.scala
```diff
@@ -1,24 +1,19 @@
 package io.scalac.mesmer.core.module
 
-import com.typesafe.config.{ Config => TypesafeConfig }
+import io.opentelemetry.instrumentation.api.config.{ Config => OTConfig }
```
lgajowy, line 3:
```suggestion
import io.opentelemetry.instrumentation.api.config.{ Config => OpenTelemetryConfig }
```

or 

```suggestion
import io.opentelemetry.instrumentation.api.config.{ Config => OtelConfig }
```

I'd prefer the non-abbreviated one - less "deciphering" on the developer side.

core/src/main/scala/io/scalac/mesmer/core/module/Module.scala
```diff
@@ -27,36 +22,32 @@ object Module {
...
+
+    fromMap(moduleConfigurations)
+  }
+
+  protected def fromMap(properties: Map[String, Boolean]): Config
 
   def defaultConfig: Result
 
-  val mesmerConfig: String = s"module.$name"
+  lazy val mesmerConfig: String = s"module.$name"
```
lgajowy, line 46: Why is this lazy?

example/src/main/resources/application.conf
```diff
@@ -30,6 +30,7 @@ akka {
     artery {
       enabled = on
       transport = tcp
+      canonical.hostname = "127.0.0.1"
       canonical.hostname = ${?HOSTNAME}
```
lgajowy, line 34:
```suggestion
      canonical.hostname = "127.0.0.1"
```

extension/src/main/scala/io/scalac/mesmer/extension/AkkaMonitoring.scala
```diff
@@ -61,13 +56,10 @@ final class AkkaMonitoring(system: ActorSystem[_])(implicit otelLoader: OpenTele
    We combine global config published by agent with current config to account for actor system having
    different config file than agent. We take account only for 4 modules as those are only affected by agent.
    */
```
lgajowy, line 58: Is this comment still accurate?

core/src/main/scala/io/scalac/mesmer/core/module/Module.scala
```diff
@@ -27,36 +22,32 @@ object Module {
...
-    def akkaActor: T
-    def akkaActorTyped: T
-  }
 }
 
 trait MesmerModule extends Module with MesmerConfigurationBase {
-  override type Result = Config
+  override type Result = Config with Product
+
+  lazy val enabled: Config = {
```
lgajowy, line 30: I'n not sure about the following: 
 - is `lazy` needed here?
 - why is it called `enabled`?  Should it be called `enable` instead (verbs indicate action)?

core/src/main/scala/io/scalac/mesmer/core/module/Module.scala
```diff
@@ -27,36 +22,32 @@ object Module {
...
+  lazy val mesmerConfig: String = s"module.$name"
 }
 
 trait MetricsModule {
   this: Module =>
   override type All[T] <: Metrics[T]
   type Metrics[T]
 }
-
-trait RegistersGlobalConfiguration extends Module {
```
lgajowy, line 54: 👍

extension/src/main/scala/io/scalac/mesmer/extension/AkkaMonitoring.scala
```diff
@@ -313,14 +291,5 @@ final class AkkaMonitoring(system: ActorSystem[_])(implicit otelLoader: OpenTele
     }
   }
 
-  /**
```
lgajowy, line 316: Thanks for doing this. That was a temporary hack not needed anymore.

otel-extension/src/main/java/io/scalac/mesmer/otelextension/MesmerAkkaPersistenceInstrumentationModule.java
```diff
@@ -1,20 +1,20 @@
...
 import io.scalac.mesmer.agent.akka.persistence.AkkaPersistenceAgent;
+
 import java.util.List;
 
 @AutoService(InstrumentationModule.class)
 public class MesmerAkkaPersistenceInstrumentationModule extends InstrumentationModule {
-  public MesmerAkkaPersistenceInstrumentationModule() {
-    super("mesmer-akka-persistence");
-  }
+    public MesmerAkkaPersistenceInstrumentationModule() {
```
lgajowy, line 12: are you using "google java format" in intelij? There's a plugin for that to make things simpler. When I was alone in the project, I kinda enforced that formatting. 🙈  Pre-commit also uses that and has good support for it. 

lmk if you have objections.

lgajowy (changes requested):
Thanks for the good stuff! Some minor questions/suggestions.

agent/src/main/scala/io/scalac/mesmer/agent/config/MesmerConfigPropertySource.scala
```diff
@@ -0,0 +1,30 @@
...
+    AkkaPersistenceModule,
+    AkkaActorSystemModule
+  )
+
+  /**
+   * Add modules has a default object that we turn to series of key-value pairs
+   * @return
+   */
+  def getProperties: util.Map[String, String] = modules.flatMap { module =>
+    module.defaultConfig.productElementNames.zip(module.defaultConfig.productIterator).map { case (key, value) =>
```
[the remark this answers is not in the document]
worekleszczy: Yeah, current approach use dots as OT automatically change all `-` and `_` into dots - I don't really agree with this approach, but it is what it is :)

agent/src/main/scala/io/scalac/mesmer/agent/akka/actor/AkkaActorAgent.scala
```diff
@@ -13,14 +13,12 @@ import io.scalac.mesmer.agent.akka.actor.impl._
...
 import io.scalac.mesmer.core.actor.ActorCellMetrics
-import io.scalac.mesmer.core.akka.version26x
-import io.scalac.mesmer.core.model._
 import io.scalac.mesmer.core.module.AkkaActorModule
 import io.scalac.mesmer.core.util.Timestamp
 
 object AkkaActorAgent
     extends InstrumentModuleFactory(AkkaActorModule)
-    with AkkaActorModule.All[AkkaActorModule.Jars[Version] => Option[Agent]]
+    with AkkaActorModule.All[Agent]
```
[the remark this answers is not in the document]
worekleszczy: It still does the same thing, only the type changed - we define interface for metrics for particular module in a trait and then we can share the same structure with `extension` etc. We could argue if it was needed even at the beginning, but I don't see point in doing so right now - let's transition completely to OT agent and see where we land - we can then discuss removing this entirely :)

core/src/main/scala/io/scalac/mesmer/core/module/AkkaActorModule.scala
```diff
@@ -53,107 +49,36 @@ object AkkaActorModule extends MesmerModule with AkkaActorMetrics with Registers
...
-        processedMessages = processedMessages,
-        failedMessages = failedMessages,
-        processingTimeMin = processingTimeMin,
-        processingTimeMax = processingTimeMax,
-        processingTimeSum = processingTimeSum,
-        processingTimeCount = processingTimeCount,
-        sentMessages = sentMessages,
-        droppedMessages = droppedMessages
+  protected def fromMap(properties: Map[String, Boolean]): AkkaActorModule.Config = {
+    val enabled = properties.getOrElse("enabled", true) // TODO should we check enabled here?
```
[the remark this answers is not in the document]
worekleszczy: I wasn't sure myself too, OK I'll remove this from modules

otel-extension/src/main/java/io/scalac/mesmer/otelextension/MesmerAkkaPersistenceInstrumentationModule.java
worekleszczy: Yeah, I use IntelliJ one, but I don't have strong feelings about java formatting - we can use formatting chosen by you :)

build.sbt
```diff
@@ -144,9 +144,11 @@ lazy val example = (project in file("example"))
       val properties = System.getProperties
 
       import scala.collection.JavaConverters._
-      for {
+      val keys = for {
         (key, value) <- properties.asScala.toList if value.nonEmpty
       } yield s"-D$key=$value"
+
+      keys
```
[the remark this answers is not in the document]
worekleszczy: Oh, changes in this PR are not required, but this thing was to enforce passing java properties passed to `sbt` to forked process. This was required then, maybe `sbt` already does it automatically right now, but I'd say this has minor priority

agent/src/main/scala/io/scalac/mesmer/agent/akka/persistence/AkkaPersistenceAgent.scala
```diff
@@ -24,67 +21,36 @@ object AkkaPersistenceAgent
...
-      persistentEventTotal = persistentEventTotalAgent.isDefined,
-      snapshot = snapshotAgent.isDefined
-    )
-
-    (resultantAgent, enabled)
-  }
-
-  def agent(config: Config): Agent = {
-    def orEmpty(condition: Boolean, agent: Agent): Agent = if (condition) agent else Agent.empty
+    // TODO can we introduce some foldable and/or functor/applicative to be able to map over those without manually checking every element here
```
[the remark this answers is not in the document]
worekleszczy: > Do you think it can be done in this PR? Maybe no need to leave a TODO?

I don't think so - I'll just remove TODO and add a ticket for this. We later on can decide if we want to spend time on this.

> We can at least do this: ...

Yeah, sure, this seems to alleviate at least some pains 🙃

lgajowy:
I was trying to run the pr with the otel agent to test the config but got the following error (for both sbt task and intelij idea run):

Sbt task: `sbt "project example" runWithOtelAgent`



```
ERROR io.opentelemetry.javaagent.OpenTelemetryAgent
java.lang.IllegalStateException: The agent was not installed, because the agent was found in '/Users/lg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/io/opentelemetry/javaagent/opentelemetry-javaagent-bootstrap/1.10.0-alpha/opentelemetry-javaagent-bootstrap-1.10.0-alpha.jar', which doesn't contain a Premain-Class manifest attribute. Make sure that you haven't included the agent jar file inside of an application uber jar.
	at io.opentelemetry.javaagent.OpenTelemetryAgent.verifyJarManifestMainClassIsThis(OpenTelemetryAgent.java:102)
	at io.opentelemetry.javaagent.OpenTelemetryAgent.installBootstrapJar(OpenTelemetryAgent.java:83)
	at io.opentelemetry.javaagent.OpenTelemetryAgent.startAgent(OpenTelemetryAgent.java:53)
	at io.opentelemetry.javaagent.OpenTelemetryAgent.premain(OpenTelemetryAgent.java:44)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:513)
	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:525)
```

Is it related to the javaagent-tooling dependency? Do you experience it as well?

worekleszczy:
@lgajowy Yeah, TBH I didn't run this with to OTEL agent, but I get this too. It seems to be more severe than I thought, I spend some time investigating this yesterday and I don't have a solution yet. We probably need to discuss this - I don't wan to be dramatical, but this might even force us to abandon switching to OTEL config in extension...

lgajowy:
@worekleszczy 

I think we should not call the ConfigInitializer.initialize() at all. It's something that the opentelemetry agent does for us in the [AgentInstaller](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/c461d22d83494ae284b3337ba3dee3985b78d068/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java#L80). I was able to confirm on my clone of this PR that when you: 
 - remove the javaagent-tooling completely and stop using the configInitializer
 - run with the otel agent 
 - call OpentelemetryConfig.get() in the `MesmerModule`
 
You still get our default values and are able to overwrite them via system props. So this is the approach I suggest for otel agent. 


If we really want to still support the Mesmer agent: 

```scala
lazy val enabled: Config = {

    val config = OpenTelemetryConfig.get()

    if (!config.getAllProperties.keySet().toArray[String].exists(_.contains("mesmer"))) {
      // build the config from scratch for the mesmer agent
      OpenTelemetryConfig
        .builder()
        .readProperties(
          MesmerConfigPropertySource.getProperties
        ) // you'd probably need to move the MesmerConfigPropertySource to the core
        .readSystemProperties()
        .readEnvironmentVariables()
    }
```

But that is imho not necessary and is just another reason to drop the mesmer agent right away and focus on the otel extension.

lgajowy:
@worekleszczy 

I think we should not call the ConfigInitializer.initialize() at all. It's something that the opentelemetry agent does for us in the [AgentInstaller](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/c461d22d83494ae284b3337ba3dee3985b78d068/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java#L80). I was able to confirm on my clone of this PR that when you: 
 - remove the javaagent-tooling completely and stop using the configInitializer
 - run with the otel agent 
 - call OpentelemetryConfig.get() in the `MesmerModule`
 
You still get our default values and are able to overwrite them via system props. So this is the approach I suggest for otel agent. 


If we really want to still support the Mesmer agent: 

```scala
lazy val enabled: Config = {

    val config = OpenTelemetryConfig.get()

    if (!config.getAllProperties.keySet().toArray[String].exists(_.contains("mesmer"))) {
      // build the config from scratch for the mesmer agent
      OpenTelemetryConfig
        .builder()
        .readProperties(
          MesmerConfigPropertySource.getProperties
        ) // you'd probably need to move the MesmerConfigPropertySource to the core
        .readSystemProperties()
        .readEnvironmentVariables()
    }
```

But that is imho not necessary and is just another reason to drop the mesmer agent right away and focus on the otel extension.

lgajowy:
@worekleszczy 

Tested the PR and compared with #344: 

I passed the following properties: 
```
-Dio.scalac.mesmer.module.akkahttp.connections=true
-Dio.scalac.mesmer.module.akkahttp.request.time=true
-Dio.scalac.mesmer.module.akkahttp.request.counter=false
```

When I put a break point in `AkkaHttpAgent` in line 42

in #328: I get config that holds 3x false for the akka HTTP agent and that results in HTTP being disabled

in #344: I get true, false, true, as expected.

The behavior in #328 doesn't seem right and I'd expect to witness the behavior from #344 in #328 (did I miss something?)

lgajowy (approved):
Seems to be working ok now. Let's decide on standup which one to merge (#344)

PR #331, opened 2022-02-25
304 akka actor metrics otel agent support

### LINKED ISSUE #304: Support Akka actor metrics in otel extension
Similar to #291 and #296.

lgajowy:
Closes #304
___

## Summary:

Instruments akka actor in otel agent. Uses `VirtualField` to store metric/timestamp fields that used to be injected in ActorCell/Envelope types by mesmer agent.


core/src/main/scala/io/scalac/mesmer/core/actor/ActorCellDecorator.scala
```diff
@@ -1,15 +1,14 @@
 package io.scalac.mesmer.core.actor
 
-import io.scalac.mesmer.core.util.ReflectionFieldUtils
+import io.opentelemetry.instrumentation.api.field.VirtualField
 
 object ActorCellDecorator {
 
-  final val fieldName = "_actorCellMetrics"
+  @inline def getMetrics(actorCell: Object): Option[ActorCellMetrics] = Option(
```
lgajowy, line 7: It was hard to find a least common type for `Cell` `ActorCell` and `ActorContext`. Before, we also operated on Object as an input parameter to `ActorCellDecorator.get()` method. 

@worekleszczy do you have any ideas on how to widen the type? If not, let's postpone this I guess. This will (should?) be simplified when we will attempt getting rid of the akka extension I guess.

core/src/main/scala/akka/MesmerMirrorTypes.scala
```diff
@@ -12,7 +12,4 @@ object MesmerMirrorTypes {
   type GraphInterpreterMirror          = GraphInterpreter
   type GraphInterpreterShellMirror     = GraphInterpreterShell
   type GraphStageLogicMirror           = GraphStageLogic
-
-  val c: Cell = ???
-  c.isLocal
```
worekleszczy, line 17: wow, we had this in repo? How this didn't fail ?🙃

agent/src/main/scala/io/scalac/mesmer/agent/Agent.scala
```diff
@@ -20,10 +20,10 @@ final case class Agent private (private[agent] val instrumentations: Set[AgentIn
 
   def ++(other: Agent): Agent = Agent(instrumentations ++ other.instrumentations)
 
-  def plus(other: Agent): Agent = ++(other)
-
   def ++(other: AgentInstrumentation): Agent = Agent(instrumentations + other)
 
+  def onCondition(condition: Boolean): Agent = if (condition) this else Agent.empty
```
worekleszczy, line 25: This is nice 🙂

agent/src/main/scala/io/scalac/mesmer/agent/akka/actor/AkkaActorAgent.scala
```diff
@@ -87,114 +79,83 @@ object AkkaActorAgent
...
+    val config = module.enabled(configuration)
+
+    List(
+      sharedInstrumentation.onCondition(config.mailboxSize),
+      (sharedInstrumentation ++ mailboxInstrumentation).onCondition(
+        config.mailboxTimeCount
+          || config.mailboxTimeSum
+          || config.mailboxTimeMin
+          || config.mailboxTimeMax
+      ),
```
worekleszczy, line 93: I'd extract all check into separate, sth like:
```
    val mailboxEnabled = config.mailboxTimeCount ||
      config.mailboxTimeSum ||
      config.mailboxTimeMin ||
      config.mailboxTimeMax

    List(
      sharedInstrumentation.onCondition(config.mailboxSize),
      (sharedInstrumentation ++ mailboxInstrumentation).onCondition(mailboxEnabled),
      ...
```

worekleszczy (changes requested):
Good job, this is good step towards removing our agent 🚀  The only think that concerns me at this point it the part where we use `VirtualField.find(classOf[Object], classOf[ActorCellMetrics]` on `Object` - I'll think about how can we change `Object` to more specific type.

agent/src/main/scala/io/scalac/mesmer/agent/akka/actor/impl/MailboxDequeueInstrumentation.scala
```diff
@@ -1,27 +1,28 @@
...
+      ActorCellDecorator.getMetrics(MailboxOps.getActor(mailbox)).foreach { metrics =>
         if (metrics.mailboxTimeAgg.isDefined) {
           add(metrics, computeTime(envelope))
         }
       }
     }
 
-  @inline private def computeTime(envelope: Object): Interval =
-    EnvelopeDecorator.getTimestamp(envelope).interval()
+  @inline private def computeTime(envelope: Envelope): Interval = EnvelopeDecorator.getInterval(envelope)
```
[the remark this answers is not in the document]
lgajowy: Why wouldn't it?

agent/src/main/scala/io/scalac/mesmer/agent/akka/actor/impl/ActorMetricsInitAdvice.scala
```diff
@@ -0,0 +1,46 @@
...
+import io.scalac.mesmer.core.actor.ActorCellDecorator
+import io.scalac.mesmer.core.actor.ActorCellMetrics
+
+object ActorMetricsInitAdvice {
+
+  @OnMethodEnter
+  def initializeMetrics(@Advice.This cell: ActorContext): Unit = {
+
+    val metrics = new ActorCellMetrics()
+    val config  = AkkaActorAgent.module.enabled(ConfigFactory.load())
```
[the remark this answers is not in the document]
lgajowy: We could use Advice.withCustomProperties().bind()`. Alternatively we can wrap the config in an object that loads it once and use a getter here to avoid loading the config multiple times. 

I will see what's best here

core/src/main/scala/io/scalac/mesmer/core/actor/ActorCellDecorator.scala
lgajowy: > This might lead to all Object having additional field if jvm even allows this

Could you post proof showing that/some more details of your investigation? I'd expect some significant performance/memory degradation if that would be the case, yet I didn't notice anything worrying.
For me now the part that is more convincing is just avoiding the Object type that could be anything. 

>  I'll take a stub at this and maybe will figure out sth.

I will do this. Above I was asking because you might have known from the top of your head.

lgajowy: That should do the trick: 22d5da2641abf48fab14e4d135e72736ed00b223

agent/src/main/scala/io/scalac/mesmer/agent/akka/actor/impl/MailboxDequeueInstrumentation.scala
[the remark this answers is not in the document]
lgajowy: Ah, for OUR agent (sorry, I'm blind). It does but it doesn't really matter anyway :)

agent/src/main/scala/io/scalac/mesmer/agent/akka/actor/AkkaActorAgent.scala
```diff
@@ -87,114 +79,83 @@ object AkkaActorAgent
...
-    )
+  lazy val mailboxTimeMax: AkkaActorModule.AkkaActorJars[Version] => Option[Agent] = commonMailboxInstrumentation
 
-  lazy val mailboxTimeCount: AkkaActorModule.Jars[Version] => Option[Agent] = versions =>
-    ifSupported(versions)(
-      sharedInstrumentation ++ mailboxInstrumentation ++ initField("mailbox-time", _.initMailboxTimeAgg())
-    )
+  lazy val mailboxTimeSum: AkkaActorModule.AkkaActorJars[Version] => Option[Agent] = commonMailboxInstrumentation
+
+  lazy val mailboxTimeCount: AkkaActorModule.AkkaActorJars[Version] => Option[Agent] = commonMailboxInstrumentation
```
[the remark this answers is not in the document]
lgajowy: The AkkaActorMetricsDef requires akkaActorAgent to implement these fields. Afaik you simplified that so we can clean it up later after merging our changes

worekleszczy (approved):
I take another look and the problem is bigger than we though - we're not generating additional field for any `Envelope` and `ActorCell`  - for both of those we're defaulting for shared state. I'm lifting requested changes as this should be resolved in another issue 🙂

PR #340, opened 2022-03-07
339 otel autoconfig

### LINKED ISSUE #339: Example: Replace Mesmer Configuration SPI with sdk-autoconfigure from opentelemetry
I found this: https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure

I think it's the simplest way of configuring mesmer and we should promote it in our example. It will still be possible to configure the application using other ways (eg. SdkBuilder). It also eliminates the need for the Opentelemetry config loading spi and afaik automatically takes care of shutting down opentelemetry sdk when needed:

> The autoconfigure module registers Java shutdown hooks to shut down the SDK when appropriate.

lgajowy:
Closes #339 
___

## Summary:

Draft PR showing example usage of opentelemetry sdk autoconfiguration. Works with otel agent 1.11.0 but i think there are some trouble in 1.10.0 version (I don't see the metrics being registered, I don't yet know why).


lgajowy:
@worekleszczy what do you think of including this one in the next release? it is replacing the SPI I developed for configuring the opentelemetry (meter provider). it is proven to be working even in the akka extension and does not require adding any code in the user's app which is nice. It also has sensible defaults (see https://github.com/ScalaConsultants/mesmer-akka-agent/issues/339) for more info and a link)

worekleszczy (approved):
LGTM - good find 🎉

PR #349, opened 2022-03-25
305 otel extension akka stream support

### LINKED ISSUE #305: Support Akka Streaming metrics in otel extension
similar to #291 and #296

worekleszczy:
Closes #305 
___

## Summary:

Adds streams support for otel agent. As we're moving away from our agent here all of the instrumentations were copied from `agent` and sometimes moved to move suitable packages, hence reviewing might get a little bit harder. We must manually add all helper classes to `application` class loader - for otel this is taken care of by `muzzle`.


example/src/main/resources/logback.xml
```diff
@@ -31,8 +31,9 @@
     <logger name="akka" level="WARN"/>
     <logger name="slick" level="INFO"/>
     <logger name="io.scalac" level="INFO"/>
+    <logger name="io.scalac.mesmer.extension.AkkaMonitoring" level="DEBUG"/>
 
-    <logger name="akka.persistence" level="TRACE" additivity="false">
+    <logger name="io.scalac.mesmer.utils" level="DEBUG" additivity="false">
```
worekleszczy, line 36: Yeah, this might not be wanted, but I don't think we need `akka.persistence` set to trace, and this one redirects logs from a little agent that logs all loaded classes and classloader that  loaded it, which is pretty useful to me 🙃

lgajowy: There're only `agent.utils` and `core.util` packages. Does this logback change work?

otel-extension/src/main/java/akka/ActorGraphInterpreterOelAdvice.java
```diff
@@ -0,0 +1,17 @@
+package akka;
+
+import akka.actor.Actor;
+import io.scalac.mesmer.agentcopy.akka.stream.impl.ActorGraphInterpreterOtelDecorator;
+import net.bytebuddy.asm.Advice;
+import scala.PartialFunction;
+import scala.runtime.BoxedUnit;
+
+public class ActorGraphInterpreterOelAdvice {
```
lgajowy, line 9:
```suggestion
public class ActorGraphInterpreterOtelAdvice {
```

otel-extension/src/main/scala/akka/ConnectionOtelOps.scala
```diff
@@ -0,0 +1,62 @@
...
+   */
+  def getAndResetCounterValues(connection: Connection): (Long, Long) = {
+    val (push, pull) = VirtualField
+      .find(classOf[Connection], classOf[(Long, Long)])
+      .get(connection)
+
+//    VirtualField
+//      .find(classOf[Connection], classOf[(Long, Long)])
+//      .set(connection, (0L, 0L))
+
```
lgajowy, line 57:
```suggestion
```

otel-extension/src/main/java/akka/stream/ConnectionConstructorAdvice.java
```diff
@@ -0,0 +1,14 @@
...
+import akka.stream.impl.fusing.GraphInterpreter;
+import io.opentelemetry.instrumentation.api.field.VirtualField;
+import net.bytebuddy.asm.Advice;
+import scala.Tuple2;
+
+public class ConnectionConstructorAdvice {
+
+    @Advice.OnMethodExit
+    public static void initCounters(@Advice.This GraphInterpreter.Connection self) {
+        VirtualField.find(GraphInterpreter.Connection.class, Tuple2.class).set(self, Tuple2.apply(0L, 0L));
```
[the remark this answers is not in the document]
lgajowy: Unless this is a Tuple2 only for legacy reasons - then you can ignore that if you don't want to refactor this

otel-extension/src/main/java/io/scalac/mesmer/otelextension/MesmerAkkaStreamInstrumentationModule.java
```diff
@@ -0,0 +1,90 @@
...
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import net.bytebuddy.matcher.ElementMatchers;
+
+import java.util.*;
+
+import static net.bytebuddy.matcher.ElementMatchers.isConstructor;
+
+@AutoService(InstrumentationModule.class)
+public class MesmerAkkaStreamInstrumentationModule extends InstrumentationModule implements InstrumentationModuleMuzzle {
```
lgajowy, line 20: Hot stuff - nice!

From the InstrumentationModuleMuzzle javadoc: 

```
/**
 * This interface contains methods that muzzle automatically adds to the {@link
 * InstrumentationModule}. They are not supposed to be used by end-users, only by our own internal
 * code.
 */
```

I suspect that this will still work with otel extension but should we worry? If for some reason this will stop being supported, will our extension still have ways to work properly? We can add the jar with helper classes to the classpath manually, right?

otel-extension/src/main/scala/akka/ActorGraphInterpreterProcessEventAdvice.scala
```diff
@@ -0,0 +1,35 @@
+package akka
+
+import _root_.io.scalac.mesmer.agentcopy.akka.stream.impl.ActorGraphInterpreterOtelDecorator
```
lgajowy, line 3: I think you can replace "agentcopy" to "otelextension.instrumentations"

lgajowy (changes requested):
Thanks! Left some comments. 

When it comes to muzzle: we need to do the same thing for other instrumentations, don't we (http, persistence)?

When it comes to copying code: if you feel the same way, given the fact that we're about to remove the mesmer agent support anyway, you can remove the old streaming rather than copy it to agantcopy. Unless you still want to be able to compare both solutions - then don't. :) Up to you.

build.sbt
```diff
@@ -220,3 +222,36 @@ def runWithOtelAgent = Command.command("runWithOtelAgent") { state =>
     Project.extract(newState).runInputTask(Compile / run, "", newState)
   s
 }
+
+def runWithOtelStream = Command.command("runWithOtelStream") { state =>
```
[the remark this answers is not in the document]
worekleszczy: Yeah, naming is always hard.

> Best option (probably out of scope of this PR): have one task that takes example variant as an input param (similar to sbt run when you have multiple main methods)

I don't think there is a need in this PR, and for now I don't think we should spend time on this. Maybe after we have some more examples to run and this become unmanageable 🙃

core/src/main/scala/akka/MesmerMirrorTypes.scala
```diff
@@ -10,6 +11,7 @@ object MesmerMirrorTypes {
   type Cell                            = akka.actor.Cell
   type ExtendedActorMaterializerMirror = ExtendedActorMaterializer
   type GraphInterpreterMirror          = GraphInterpreter
+  type ConnectionMirror                = Connection
```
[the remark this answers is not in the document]
worekleszczy: Yeah, unfortunately not my invention, I think I spotted this somewhere 😛

example/src/main/resources/logback.xml
worekleszczy: Yeah, this is for something outside of the current repo - https://github.com/worekleszczy/monitor-agent

otel-extension/src/main/java/akka/stream/ConnectionConstructorAdvice.java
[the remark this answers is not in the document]
worekleszczy: Yeah, tbh I went for `Tuple2` just to see if this works, we can introduce something custom, as least not to use generics for this.

otel-extension/src/main/java/io/scalac/mesmer/otelextension/MesmerAkkaStreamInstrumentationModule.java
worekleszczy: `getMuzzleHelperClassNames` is part for the `InstrumentationModule` api, so we're safe. The only functionality that we use from `InstrumentationModuleMuzzle` is :
```
@Override
    public void registerMuzzleVirtualFields(VirtualFieldMappingsBuilder builder) {
        builder.register("akka.stream.impl.fusing.GraphInterpreter$Connection", "scala.Tuple2");
    }
```
And it's to make opentelemetry actualy turn virtual field into actual field on the class

otel-extension/src/main/scala/akka/ActorGraphInterpreterProcessEventAdvice.scala
worekleszczy: Yeah, I chose the name just to have something different than `agent ` to be sure we load the correct classes 🙂 we can move it to `otelextension.instrumentations`

example/src/main/resources/logback.xml
lgajowy: isn't java's `-verbose:class` option enough?

lgajowy: I agree that we should not trace akka persistence, so I suggest removing both:

```suggestion
```

lgajowy: I also checked if java -verbose:class is enough but it does not seem to show you the classloader

worekleszczy: I agree that we can remove this here . I don't think `-verbose:class` is enough for my use case - I tried using this before I created another agent and had a hard time redirecting it to another file. This should be possible with new java9 syntax for logs, but I didn't succeed in doing it.

worekleszczy:
> When it comes to muzzle: we need to do the same thing for other instrumentations, don't we (http, persistence)?

Yes, this is required. Here is the ticket for it #354 

> When it comes to copying code: if you feel the same way, given the fact that we're about to remove the mesmer agent support anyway, you can remove the old streaming rather than copy it to agantcopy. Unless you still want to be able to compare both solutions - then don't. :) Up to you.

For now I'll leave those in - we're removing `agent` entirely, so it won't be a problem for long

PR #358, opened 2022-03-31
Remove max value grafana graphs (not working) and cleanup the dashboards

### LINKED ISSUE #297: reinstate/modify/remove max metric values in Grafana dashboard
After bumping the opentelemetry version to version > 1.6 we no longer have a value recorder. It was replaced with histograms. The problem is that we do not interpret the data provided by the histograms properly in the grafana dashboard. This is why all the "max" value graphs show "no data". 

Goal: investigate what should we do. Should we: 
 - remove the max graphs?
 - (or) fix them somehow (show the whole histograms? 

Fixing/removing the graphs is part of the solution.

lgajowy:
Closes #297 
___

## Summary:

Removes the max dashboards. There's no way to get MAX value from histograms.

### DIFF
--- a/example/docker/dashboards/akka-cluster.json
+++ b/example/docker/dashboards/akka-cluster.json
@@ -1,19 +1,4 @@
 {
-  "__inputs": [],
-  "__requires": [
-    {
-      "id": "grafana",
-      "name": "Grafana",
-      "type": "grafana",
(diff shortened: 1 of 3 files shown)


worekleszczy (approved):
LGTM 🎉

PR #359, opened 2022-04-01
354 helper classes

### LINKED ISSUE #354: Mark classes used by Advices as helper classes in InstrumentationModules
For `akka-stream` module we copied advices files into `otel-extension` project - we should do the same for other modules and mark those as helpers to instruct `openTelemetry` to inject those in application class loader. This is also a good moment to manually add `muzzle` integration to report `VirutalField`s in use.

worekleszczy:
Closes #354
___

## Summary:

TBD

### DIFF
--- a/agent/src/main/scala/io/scalac/mesmer/agent/akka/persistence/impl/RecoveryCompletedAdvice.scala
+++ b/agent/src/main/scala/io/scalac/mesmer/agent/akka/persistence/impl/RecoveryCompletedAdvice.scala
@@ -19,7 +19,6 @@ object RecoveryCompletedAdvice {
     @Argument(0) actorContext: ActorContext[_],
     @This self: AnyRef
   ): Unit = {
-    println("Recovery completed agent")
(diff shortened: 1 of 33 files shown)


lgajowy:
Just a heads up related to the conversation we had on standup: I took a look and I see that when we try to run the tests with the new AkkaStreamAgent, we get an NPE due to an empty VirtualField. This in turn is due to the fact that `ConnectionConstructorAdvice` is not picked up during testing. It was not detected in https://github.com/ScalaConsultants/mesmer/pull/349 because the tests were not moved together with the new implementation - Github Actions was testing the old code without changes.

I'm not 100% sure if we need to do something with the dependencies (we talked about provided scope being a problem). Maybe we should fix the AkkaStreamAgent to pass the tests or modify the `InstallModule` utility.  

The `ConnectionConstructorAdvice` seems to be picked up during running the extension with otelAgent (the streaming instrumentations work and there's no NPE). 

I will continue investigating next week.

PR #362, opened 2022-04-06
351 remove akka http

### LINKED ISSUE #351: Remove akka http metrics
We don't need them anymore. Akka http instrumentation from otel agent: 
 - supports the same akka HTTP versions as our instrumentation
 - has active_requests metic that is analogous to or connection_count
 - has http_server_request duration histogram - it does the same as our request duration histogram
 - attributes are the same

lgajowy:
Closes #351 
___

## Summary:

- Removes akka http metrics
- uses openTelemetry's akka http metrics in grafana dashboards
- fixes docs


lgajowy:
There's one caveat I missed when creating the issue for that. I thought that opentelemetry metrics will have http_target attribute that's equivalent (or even more specific) to our http_path. However, although available, it seems to not be included deliberately due to high cardinality (makes total sense).  

Options I see now: 

1. still remove our metrics. opentelemetry metrics still are valuable and in the future we could extend them with our custom attributes and probably even build on the existing otel instrumentation ([by just providing extra attributes](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md#i-want-to-edit-some-attributes-and-their-values-depend-on-a-specific-db-connection-instance)). Or alternatively we could write something alongside but without using the otel extension  (I tested this [here](https://github.com/ScalaConsultants/mesmer/discussions/356))

2.  Just move the metrics to otel-extension module. The drawback is that 90% of akka-http functionality will be doubled (same data in both otel and mesmer instrumentations)

@worekleszczy opinion? I'd still go with "1"

lgajowy:
We discussed this on today's standup with @adamgadomski and @jczuchnowski. The plan for this issue would be: 
1. Merge this removal pr
2. First items to take after the release would be: https://github.com/ScalaConsultants/mesmer/issues/365 and https://github.com/ScalaConsultants/mesmer/issues/366

CONTRIBUTORS.md
```diff
@@ -10,8 +10,8 @@ Projects is comprised of 3 major parts:
 - application that is being instrumented - in repo we have the "example" app, but this in theory this could be any application that uses akka.
 
 The most interesting parts are java agent and akka extension. They have to work in tandem - `agent` creates a foundation on which `extension` can build and calculate metrics.
-It's good to use an example to get a better understanding - `agent` allows to trigger an event whenever http request starts and completes
-We've found places where those events should be triggered by looking at Akka source code and then modified necessary classes during load). 
+It's good to use an example to get a better understanding: `agent` allows to trigger an event whenever akka persistence query starts and completes.
```
worekleszczy, line 13: wow, I wouldn't remember about it 🙃

PR #367, opened 2022-04-07
354 move persistency to otel extension

### LINKED ISSUE #354: Mark classes used by Advices as helper classes in InstrumentationModules

worekleszczy:
Closes #354 
___

## Summary:

Moves all `akka-persistence` instrumentation into `otel-extension` project and mark all required helper classes for opentelemetry


lgajowy (approved):
:100: Nice! Made some minor amendments (see above commits fyi) but nothing crucial

PR #388, opened 2022-04-20
299 speed up pre commit

### LINKED ISSUE #299: Find a way for a pre-commit hook to run faster (research)
Currently the [pre-commit](https://pre-commit.com/) hook is quite slow due to running sbt formatter and scalafix. It would be super useful to make it faster. 

Options to consider: 
 - run sbt natively somehow
 - use nailgun to reduce the jvm overhead: https://github.com/facebook/nailgun
 - ???

lgajowy:
Closes #299 
___

## Summary:

- encorurages using sbt native client locally
- uses sbt native client in github actions

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -2,12 +2,15 @@ name: Scala CI
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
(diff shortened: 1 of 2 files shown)


worekleszczy (approved):
This is a very good finding 🎉 - I was not aware of this at all 🙃

PR #395, opened 2022-04-25
Bump otel version to 1.13.1*. Fix compilation errors

### LINKED ISSUE #394: Update Opentelemetry to newest version
Creating a non-scala-steward issue because, judging from otel's release notes, there might be breaking changes. Other than that it's best to update all otel dependencies together. 

Current otel version: 1.13.0

Extra work: we can use the opportunity to cleanup the dependencies. It seems that we can just use opentelemetry-api dependency that will provide API for metrics (thus we don't need opentelemetry-api-metrics).

lgajowy:
*1.13.0 for sdk-autoconfigure - that's the latest for this one.

Closes #394 
___

## Summary:

- bumps to latest version
- removes unnecessary dependencies 
- removes unnecessary configuration (-Dotel.metric.exporter is now otlp by default)
- fixes compilation errors

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -54,12 +54,9 @@ lazy val core = (project in file("core"))
     name := "mesmer-akka-core",
     libraryDependencies ++= {
       akka ++
-      openTelemetryApi ++
-      openTelemetryApiMetrics ++
       openTelemetryInstrumentation ++
       scalatest ++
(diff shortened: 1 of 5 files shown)


worekleszczy (approved):
LGTM, I run it a couple of times and no issues with `sbtn`

PR #400, opened 2022-05-02
376 migrate tests to otel agent installer

### LINKED ISSUE #376: Migrate tests to opentelemetry agent, not custom bytebuddy agent installer
Currently, we use the latter, we should use the former. Only this way we will be covering the same scope in tests as actually running the instrumentations. 

For example we currently do not cover in testing the InstrumentationModule classes that register the VirtualFields. So the virtual fields in tests are weak, concurrent hashmaps underneath rather than being attached to the object that "has" the virtual field.

lgajowy:
Closes #376 
___

## Summary:

- runs integration tests with OTEL agent 
- moved the tests that require the agent to `it` directory
- have a separate task for running integration tests


.github/workflows/scala.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       with:
         java-version: 1.8
     - name: Run tests
-      run: sbt test
+      run: sbt clean; sbt testAll
```
lgajowy, line 39: I struggled to define a dependency between IntegratonTest/test task and assembly that would force running assembly and refreshing the artifacts. Therefore, it's possible that a stale jar will be used for testing unless we run "clean" before running the tests. 

If there's a way to force the dependency so that the IntegrationTest/test task always reassembles the extension jar, I'd be happy to incorporate it.

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/AgentInstaller.scala
```diff
@@ -0,0 +1,37 @@
...
+import io.scalac.mesmer.agent.AgentInstrumentation
+import io.scalac.mesmer.agent.util.i13n
+
+trait AgentInstaller {
+  def install(agent: Agent): Unit
+}
+
+object AgentInstaller {
+
+  def make(agentBuilder: AgentBuilder, instrumentation: Instrumentation): AgentInstaller =
```
worekleszczy, line 17: Do we need this?

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/OtelAgentTest.scala
```diff
@@ -0,0 +1,21 @@
...
+import io.opentelemetry.context.ContextStorage
+import io.opentelemetry.instrumentation.testing.AgentTestRunner
+import io.opentelemetry.instrumentation.testing.util.ContextStorageCloser
+import org.scalatest.BeforeAndAfterAll
+import org.scalatest.BeforeAndAfterEach
+import org.scalatest.TestSuite
+
+abstract class OtelAgentTest extends TestSuite with BeforeAndAfterAll with BeforeAndAfterEach {
+
+  private val testRunner = AgentTestRunner.instance()
```
worekleszczy, line 12: I sense that this is the most crucial part of this PR, right? It still should run a test suit per process? Previously we limited amount of test run in the separate process to allow any suite to register it's transformations and not intervene with previously registered - is it still the case?

worekleszczy (approved):
Overall looks really good, can't wait to use it 🎉

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/OtelAgentTest.scala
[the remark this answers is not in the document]
lgajowy: As far as I remember it can use one agent for multiple suites. It does not pick which instrumentations are ones are installed - all are installed at the same time. Imho that's good - in case we have any conflicts/problems due to this, we certainly would have them outside testing. So it's good to detect such scenarios here, if any (didn't spot any problems yet).

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/AgentInstaller.scala
lgajowy: nope

PR #402, opened 2022-05-05
366 path attribute updated

### LINKED ISSUE #366: Add akka http path attribute to opentelemetry http metrics
The Path attribute was there in the old Akka HTTP Mesmer metrics. Opentelemetry metrics do not use it. We can extend the existing open telemetry Akka HTTP metrics by providing the attribute bu either: 
 - writing an extension that would add the attribute to [existing otel's metrics](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md#i-want-to-edit-some-attributes-and-their-values-depend-on-a-specific-db-connection-instance) in context
 - writing the Akka HTTP metrics on our own but this time using the way open telemetry proposes (without using the Mesmer's Akka extension, as proposed here: https://github.com/ScalaConsultants/mesmer/discussions/356

We'd ideally like to avoid having too great a path attribute cardinality. Therefore, a configuration mechanism can be added that will let users specify which paths to track. For example, we could use some regexp matchers: 

```-Dakka.http.paths.to.track="/foo/bar/[0-9]/.*foo/"```

The following matcher would register metrics for `foo/bar/1/afoo` path but would not register `/foo/bar/baz/afoo`.

### PR DESCRIPTION
**Pull request by lgajowy**
Closes #366 
___

## Summary:

- reintroduces http_route attribute.

### CODE REVIEW

**Comment by lgajowy** [comment]
(opened a PR so that I could leave review comments. I will do it  tomorrow)

File: otel-extension/src/main/java/io/scalac/mesmer/instrumentation/http/impl/RequestContextImplCopyAdvice.java [hunk 1]
```diff
@@ -0,0 +1,17 @@
...
+
+public class RequestContextImplCopyAdvice {
+
+  @Advice.OnMethodExit
+  public static void onExit(
+      @Advice.Return(readOnly = false) RequestContext nextContext,
+      @Advice.This RequestContext self) {
+
+    VirtualField.find(RequestContext.class, String.class)
+        .set(nextContext, VirtualField.find(RequestContext.class, String.class).get(self));
```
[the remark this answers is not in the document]
worekleszczy: This is no longer needed :)

core/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/http/NamedPathMatcher.scala
```diff
@@ -0,0 +1,19 @@
+package io.scalac.mesmer.otelextension.instrumentations.akka.http
+
+import akka.http.scaladsl.model.Uri
+import akka.http.scaladsl.server.PathMatcher
+
+trait NamedPathMatcher {
```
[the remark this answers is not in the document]
worekleszczy: This is no longer needed :)

core/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/http/NamedPathMatcher.scala
```diff
@@ -0,0 +1,19 @@
...
+
+  def pathName: String
+}
+
+final class NamedPathMatcherWrapper[T](private val underlying: PathMatcher[T], name: String)
+    extends PathMatcher[T]()(underlying.ev)
+    with NamedPathMatcher {
+
+  def pathName: String = name
```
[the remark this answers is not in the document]
worekleszczy: I don't think this is needed - we don't make this a public api, so nobody should really use it

example/src/main/scala/example/api/AccountRoutes.scala
```diff
@@ -35,45 +35,57 @@ class AccountRoutes(
...
           }
-        }
+        } ~
+        (pathPrefix("poligon") {
+          (pathPrefix(DoubleNumber | JavaUUID)) { _ =>
+            complete(StatusCodes.OK)
+          } ~ (pathPrefix(matcher | matcher2) & pathEndOrSingleSlash) { (_, _, _) =>
+            complete(StatusCodes.OK)
+          }
+        })
```
worekleszczy, line 81: delete

lgajowy: is it to be deleted?

otel-extension/src/main/java/io/scalac/mesmer/instrumentation/http/impl/HexLongNumberTemplateAdvice.java
```diff
@@ -0,0 +1,13 @@
+package io.scalac.mesmer.instrumentation.http.impl;
+
+import akka.http.scaladsl.server.PathMatcher;
+import io.opentelemetry.instrumentation.api.field.VirtualField;
+import net.bytebuddy.asm.Advice;
+
+public class HexLongNumberTemplateAdvice {
```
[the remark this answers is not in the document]
lgajowy: I think you could  add method specific parameters and bind them to advice. Something like this: 
```Advice.withCustomConfiguration().bind(...).to(MyAdvice.class).```

File: otel-extension/src/main/java/io/scalac/mesmer/instrumentation/http/impl/SlashPathTemplateAdvice.java [hunk 6]
```diff
@@ -0,0 +1,13 @@
+package io.scalac.mesmer.instrumentation.http.impl;
+
+import akka.http.scaladsl.server.PathMatcher;
+import io.opentelemetry.instrumentation.api.field.VirtualField;
+import net.bytebuddy.asm.Advice;
+
+public class SlashPathTemplateAdvice {
```
lgajowy, line 7: (similar to above) looks like all (most of) template advices could have a generic advice that is parametrized

otel-extension/src/main/java/io/scalac/mesmer/otelextension/LoggingAgentExtension.java
```diff
@@ -0,0 +1,65 @@
...
+import io.opentelemetry.javaagent.tooling.AgentExtension;
+import java.io.File;
+import java.io.IOException;
+import net.bytebuddy.agent.builder.AgentBuilder;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.dynamic.DynamicType;
+import net.bytebuddy.utility.JavaModule;
+
+@AutoService(AgentExtension.class)
+public class LoggingAgentExtension implements AgentExtension {
```
lgajowy, line 13: Please remove it. It seems not to be the part of the issue solution. 

If we want to have util like this: 
 - it should be managed independently from the library instrumentations. For example a separate otel agent extension that has to be deliberately turned on
 - it has to be parametrized so that there are no "piotrjosiak" file paths ;). What else should be parametrized here?
 - it needs documentation, such as running and configuration instructions. We will have some junior devs pretty soon - for sure this piece will be enigmatic 

So I suggest to remove it now and have it incorporated in another issue where we can focus on this feature and ✔️ all the bullets above

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/http/Instrumentation.scala
```diff
@@ -0,0 +1,196 @@
...
+import net.bytebuddy.asm.Advice
+import net.bytebuddy.description.`type`.TypeDescription
+import net.bytebuddy.description.method.MethodDescription
+import net.bytebuddy.implementation.SuperMethodCall
+import net.bytebuddy.matcher.ElementMatcher
+import net.bytebuddy.matcher.ElementMatchers
+
+import io.scalac.mesmer.instrumentation.http.impl.OverridingRawPathMatcher
+
+object Instrumentation {
```
lgajowy, line 14: maybe rename to "PathAttributeExtractors" (or similar)? "Instrumentation" to me sounds like you will attach some "instruments" here. We just add advice with VirtualFields setters/getters. WDYT? 

Also might be confusing when we add a new HTTP instrumentation (connections counter that is pending)

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/http/OverridingRawPatchMatcherImpl.scala
```diff
@@ -0,0 +1,55 @@
...
+import akka.http.javadsl.server.Rejected
+import akka.http.scaladsl.server.Directive
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.PathMatcher
+import akka.http.scaladsl.server.PathMatcher.Matched
+import akka.http.scaladsl.server.PathMatcher.Unmatched
+import io.opentelemetry.context.Context
+import io.opentelemetry.instrumentation.api.field.VirtualField
+
+object OverridingRawPatchMatcherImpl {
```
lgajowy, line 12: it's very difficult to understand what this class does. Do you think you could document that? What exactly is a "raw Path"?

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/http/UpdateHttpRouteWrapper.scala
```diff
@@ -0,0 +1,33 @@
...
+) extends (HttpRequest => Future[HttpResponse]) {
+  def apply(v1: HttpRequest): Future[HttpResponse] = {
+
+    val wrapper = new RouteTemplateHolder
+    val newContext = Context
+      .current()
+      .`with`(RouteContext.routeKey, wrapper)
+
+    Using.resource(newContext.makeCurrent()) { _ =>
+      println("!+!+!+!+!+!+!+!+!+!+! Wrapping a request !+!+!+!+!+!+!+!+!+!+!")
```
lgajowy, line 24: TO BE REMOVED

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/http/UpdateHttpRouteWrapper.scala
```diff
@@ -0,0 +1,33 @@
...
+import akka.http.scaladsl.model.HttpResponse
+import io.opentelemetry.context.Context
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpRouteHolder
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpRouteSource
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.util.Using
+
+final class UpdateHttpRouteWrapper(inner: HttpRequest => Future[HttpResponse])(implicit
```
lgajowy, line 13: Can this simply be called "HttpRouteUpdater"?

lgajowy (commented):
That great work! Some comments though - I think this can be cleaned up significantly with bytebuddy's parameter binding. There are also some rough edges and debugging leftovers. I think this PR is really great though!

Since I created the PR I cant "request changes" but please treat this review this way.

One final note: the tests fail - no idea why. I can't reproduce that locally. Do you experience any problems? Possible reasons from the top of my head: 
- sbt native is freaking out (we use it in github actions)
- it's the "piotrjosiak" path in the code that unreachable :D

example/src/main/scala/example/api/AccountRoutes.scala
worekleszczy: Yeah

otel-extension/src/main/java/io/scalac/mesmer/instrumentation/http/impl/NumberTemplateAdvice.java
```diff
@@ -0,0 +1,13 @@
+package io.scalac.mesmer.instrumentation.http.impl;
+
+import akka.http.scaladsl.server.PathMatcher;
+import io.opentelemetry.instrumentation.api.field.VirtualField;
+import net.bytebuddy.asm.Advice;
+
+public class NumberTemplateAdvice {
+
+  @Advice.OnMethodExit
+  public static void onExit(@Advice.This PathMatcher<?> self) {
```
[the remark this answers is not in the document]
worekleszczy: Yes, this is a good improvement 🙂

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/http/UpdateHttpRouteWrapper.scala
worekleszczy: I think that `Wrapper` is a valid name, as it reflects what this class actually does 🙂

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/http/Instrumentation.scala
worekleszczy: Hmm, I see it as a collection of `TypeInstrumentations`, so I'g go with `PathMatchingInstrumentations` or something similar

otel-extension/src/main/java/io/scalac/mesmer/instrumentation/http/impl/HexLongNumberTemplateAdvice.java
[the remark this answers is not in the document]
worekleszczy: I guess that It's not needed here, as we don't really extract this number, but only want to match a type

worekleszczy (approved):

PR #404, opened 2022-05-11
365 akka http connection metrics

### LINKED ISSUE #365: Reinstate Akka http connections metric
Issue https://github.com/ScalaConsultants/mesmer/issues/351 removed akka http and encouraged using opentelemetry akka http metrics instead. These metrics do not contain akka http conenction. However, we can write an extension in mesmer that reinstates that metric using a new implementation and not relaying on the mesmer akka extension. A similar thing was previously done in a PoC for this proposal: https://github.com/ScalaConsultants/mesmer/discussions/356

lgajowy:
Closes #365 
___

## Summary:

- adds http connections metric with it's config
- adds grafana dashboard for it


lgajowy:
@worekleszczy Some unit/integration testing is still needed - i need to figure out how to do it best

lgajowy:
@worekleszczy Regarding adding testing to this pr - it seems to be a bigger research than I initially anticipated:

- maybe it's possible to use the context similar to what you did in pr #402. if we can use otel context to assert the fact that some connection-related attributes appeared, that could lead us somewhere. Sounds a little bit hacky in this particular case, though. 

- it would be a lot better to be have some test MetricsSDK used by the AgentForTesting. The instrument instances (eg. connections counter) would assert if the value was passed to them in the test. No idea if this is doable.

- I noticed that there are methods like below in the AgentTestRunner: 

```
testRunner.runWithClientSpan()
testRunner.runWithServerSpan()
```

this looks like something useful to assert if the span was created correctly. We do not use spans however but maybe there's something similar for metrics alone?

PR #411, opened 2022-05-31
383 actor path

### LINKED ISSUE #383: Make Actor Path configurable in akka Actor metrics
Currently the actor_path attribute in akka actor metrics is very limited. It's either "/user" or "system" which is not very specific. 

Options to consider: 
 - use full actor path as an attribute. Will it be safe performance-wise? Won't the high cardinality of different attribute set affect the performance?
  - have a way of filtering which paths to track, eg using `path.contains("/user/foo") or a regex 

Similar issues: 
https://github.com/ScalaConsultants/mesmer/issues/366 where there's a similar attribute (http_path) that faces the same problem with attributes

worekleszczy:
Closes #383  
___

## Summary:

Fixes how we compute grouping of actors and add new operation to configuration


lgajowy:
@worekleszczy please merge if you think the same

PR #418, opened 2022-06-22
Feature/remove akka actor from extension

### LINKED ISSUE #384: Move away from akka extension in akka Actor metrics
The idea is to stop using Mesmer Akka Extension for Akka Actor metrics. We can propagate the all contextual data in the "Context" object that is saved for each ActorCell and/or Envelope object we create. The context can contain Attributes set that could be later used in an opentelemetry instrument to report the metric.

This way we can resign from propagating events using an EventBus from the akka extension and hence stop using it. 

The thread here: https://github.com/ScalaConsultants/mesmer/discussions/356 shows more context and has links for the PoC where the approach was tried.

worekleszczy:
Closes #384
___

## Summary:

Injects otel instruments into actors system along with configuration necessary to export metrics from actorCell. From now on we won't rely on extension to export metrics 🎉


lgajowy (changes requested):
I stopped at ActorCellReceivedAdvice. Will continue tomorrow. Feel free to start answering/fixing - I will probably have more than 2 iterations here. Thanks!

core/src/main/scala/io/scalac/mesmer/core/actor/ActorRefConfiguration.scala
```diff
@@ -0,0 +1,38 @@
+package io.scalac.mesmer.core.actor
```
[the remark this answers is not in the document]
worekleszczy: Yeah, it might be, initially I put it there to make it slightly easier while developing

otel-extension/src/main/java/akka/actor/impl/ActorCellInitAdvice.java
```diff
@@ -1,13 +1,29 @@
...
   public static void initActorCell(
-      @Advice.FieldValue(value = "_actorCellMetrics", readOnly = false) ActorCellMetrics metrics) {
-    metrics = new ActorCellMetrics();
+      @Advice.This ActorContext self,
+      @Advice.FieldValue("system") ClassicActorSystemProvider system) {
+    Instruments instruments =
+        VirtualField.find(ActorSystem.class, Instruments.class).get(system.classicSystem());
+
+    if (instruments != null) {
+      Attributes attributes = instruments.config().forClassic(self.self()).build();
```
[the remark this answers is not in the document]
worekleszczy: I don't know if understand this correctly, but my reasoning is:
* there might be many configurations possible for an actor system, we probably will create it from configuration file
* we cache this configuration inside `ActorSystem` 
* later on we use said configuration to create a set of `Attributes` for an `ActorCell` and cache it in the virtual field

I'm not sure if we can achieve all of the above with what you propose

core/src/main/scala/io/scalac/mesmer/core/util/Timestamp.scala
```diff
@@ -10,6 +10,10 @@ final class Interval(private val nanos: Long) extends AnyVal {
   def toNano: Long   = nanos
 }
 
+object Interval {
+  def toMillis(nanos: Long): Long = math.floorDiv(nanos, 1_000_000)
```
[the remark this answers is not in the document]
worekleszczy: Reason for this is that we use `Interval` as a value class in the scala code, but java has no way of optimising value classes, so not to create intermediate representation I added a new method. For now I would just flag it and then after migrating `persistency` we'll see where we are and do major cleaning up PR

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/OtelAgentTest.scala
```diff
@@ -1,21 +1,87 @@
...
+
+  protected val testRunner = AgentTestRunner.instance()
 
   override protected def beforeAll(): Unit = testRunner.beforeTestClass()
 
   override protected def afterAll(): Unit = testRunner.afterTestClass()
 
-  override protected def beforeEach(): Unit = testRunner.clearAllExportedData()
+  override protected def beforeEach(): Unit = {
+    super.beforeEach()
```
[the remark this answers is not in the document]
worekleszczy: This is a pattern and a convention to always build `beforeEach` and `afterEach` like that - it uses stackable trait pattern, so it depends on the order of traits that we add to a final class- to keep all other modules that could use `beforeEach` we have to call `super` here to not get surprised later

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/OtelAgentTest.scala
```diff
@@ -1,21 +1,87 @@
...
+import scala.util.Try
 
-  private val testRunner = AgentTestRunner.instance()
+trait OtelAgentTest extends TestSuite with BeforeAndAfterAll with Eventually with BeforeAndAfterEach with OptionValues {
+
+  protected val testRunner = AgentTestRunner.instance()
 
   override protected def beforeAll(): Unit = testRunner.beforeTestClass()
 
   override protected def afterAll(): Unit = testRunner.afterTestClass()
```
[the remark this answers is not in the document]
worekleszczy: Hmm, I don't remember removing it TBH but I must say thet this Runner didn't execute any of the `beforeEach` and `afterEach` logic anyway, so maybe I have removed it 🙈 We can discuss what was going on with those test in a call, cause this was a mindbending exercise (and yes, it was stackable trait pattern in the end :P )

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/OtelAgentTest.scala
```diff
@@ -1,21 +1,87 @@
...
+
+    val lowestBoundary  = math.max(0.0, boundary - toleration)
+    val highestBoundary = boundary + toleration
+
+    val lowestIndex = point.getBoundaries.asScala.toVector.zipWithIndex.collectFirst {
+      case (boundary, index) if boundary >= lowestBoundary => index
+    }.getOrElse(0)
+
+    val highestIndex = point.getBoundaries.asScala.toVector.zipWithIndex.collectFirst {
+      case (boundary, index) if boundary >= highestBoundary => index
```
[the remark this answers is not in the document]
worekleszczy: For lowest boundary we use `case (boundary, index) if boundary >= lowestBoundary`. But this questions show that we really need a proper test for this method 🙈

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/OtelAgentTest.scala
```diff
@@ -1,21 +1,87 @@
...
+    }
+
+  /*
+    This might make your test flaky! Remember to adjust toleration for CI.
+
+    Here we calculate which buckets counts should we take into consideration depending on boundary and
+    toleration - count in each bucket mean that an action took LESS that this counter boundary. We
+    take toleration into the mix to mitigate flakiness when things start to take longer than anticipated
+   */
+  // TODO add tests for this
```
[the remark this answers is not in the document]
worekleszczy: I'll create some tests 🙂

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/actor/Instruments.scala
```diff
@@ -0,0 +1,47 @@
+package io.scalac.mesmer.otelextension.instrumentations.akka.actor
+import io.opentelemetry.api.metrics.LongCounter
+import io.opentelemetry.api.metrics.LongHistogram
+import io.opentelemetry.api.metrics.MeterProvider
+
+import io.scalac.mesmer.core.actor.ActorRefConfiguration
+
+// TODO can we add resources here to automatically inject actor system tags?
+final class Instruments(val config: ActorRefConfiguration, provider: MeterProvider) {
```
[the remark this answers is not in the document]
worekleszczy: Yeah, TBH I did a small cleaning up in metrics by removing `received_messages` as this information is in the `processing_time` histogram, but I'm not 100% if it was a good call :/ We should discuss this

core/src/main/scala/io/scalac/mesmer/core/actor/ActorRefConfiguration.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import io.scalac.mesmer.core.akka.model.AttributeNames
+
+trait ActorRefConfiguration {
+
+  def forClassic(ref: ClassicActorRef): AttributesBuilder
+  def forTyped(ref: ActorRef[_]): AttributesBuilder
+}
+
+object DefaultActorRefConfiguration extends ActorRefConfiguration {
```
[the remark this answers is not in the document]
lgajowy: So the high level flow of work would be: 
- merge 411 first
- finish up 418 and remove all unnecessary bits of Actor Metrics from the Akka Extension (reducing the codebase)
- migrate akka persistence to not use the akka extension as well
- then work on improving the path matching

wdyt?

lgajowy (commented):
Some new comments after reviewing both PRs (411 & 418)

otel-extension/src/main/java/akka/actor/impl/typed/SupervisorHandleReceiveExceptionAdvice.java
```diff
@@ -0,0 +1,46 @@
...
+        VirtualField.find(ActorContext.class, Attributes.class).get(classicContext);
+    Instruments instruments =
+        VirtualField.find(ActorSystem.class, Instruments.class).get(classicContext.system());
+
+    if (Objects.nonNull(instruments) && Objects.nonNull(state) && Objects.nonNull(attributes)) {
+      state.setFailed();
+      instruments.failedMessages().add(1L, attributes);
+    } else {
+      // this should never happen
+      classicContext
```
lgajowy, line 40: The error message is only about actor cell state but will be printed in case: 
 - instruments are null
 - attributes are null
 - the state is null
 
 Please fix the message appropriately. Also please consider making it an "error" since this is a very unwelcome state that should not be ignored

lgajowy: Actually, given the fact that the instruments, attributes and state should never be null ("this should neve happen"), I wonder if I'm not even ok with them throwing NPE, so no if/else. Should such an impossible state happen, that would mean that rapid reaction is needed. Logs will be ignored. NPE will not. :)

lgajowy: (the same applies to all advice)

otel-extension/src/main/scala/io/scalac/mesmer/agent/util/dsl/Model.scala
```diff
@@ -0,0 +1,3 @@
+package io.scalac.mesmer.agent.util.dsl
+
+object Model {}
```
lgajowy, line 3: Is this used anywhere?

lgajowy (changes requested):
Posted another batch of comments, please take a look. Thanks!

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/OtelAgentTest.scala
```diff
@@ -1,21 +1,87 @@
...
+   */
+  protected def assertMetrics(instrumentationName: String, metricName: String, successOnEmpty: Boolean)(
+    testFunction: PartialFunction[MetricData, Unit]
+  ): Unit =
+    eventually {
+      val result = testRunner.getExportedMetrics.asScala
+        .filter(data => data.getInstrumentationScopeInfo.getName == instrumentationName && data.getName == metricName)
+        .collect {
+          // this will be executed once for every exported services - roughly once for every 100ms of test duration
+          case data if testFunction.isDefinedAt(data) => Try(testFunction.apply(data))
```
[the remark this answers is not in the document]
lgajowy: After using it myself I see what the point is here. Let's leave this as is

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/OtelAgentTest.scala
```diff
@@ -1,21 +1,87 @@
...
+    This might make your test flaky! Remember to adjust toleration for CI.
+
+    Here we calculate which buckets counts should we take into consideration depending on boundary and
+    toleration - count in each bucket mean that an action took LESS that this counter boundary. We
+    take toleration into the mix to mitigate flakiness when things start to take longer than anticipated
+   */
+  // TODO add tests for this
+  protected def getBoundaryCountsWithToleration(
+    point: HistogramPointData,
+    boundary: Double,
```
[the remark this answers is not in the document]
worekleszczy: Yeah, it should do it :)

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/OtelAgentTest.scala
```diff
@@ -1,21 +1,84 @@
...
+    super.beforeEach()
+    testRunner.clearAllExportedData()
+  }
+
+  protected def assertMetrics(metricName: String, successOnEmpty: Boolean = false)(
+    testFunction: PartialFunction[MetricData, Unit]
+  ): Unit = assertMetrics("mesmer", metricName, successOnEmpty)(testFunction)
+
+  /**
+   * @param instrumentationName
```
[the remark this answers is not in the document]
worekleszczy: I won't apply change as I modifies this api in #440 🙂

otel-extension/src/main/java/akka/actor/impl/BoundedQueueBasedMessageQueueQueueAdvice.java
```diff
@@ -0,0 +1,17 @@
+package akka.actor.impl;
+
+import akka.dispatch.BoundedQueueBasedMessageQueue;
+import io.opentelemetry.instrumentation.api.field.VirtualField;
+import java.util.concurrent.BlockingQueue;
+import net.bytebuddy.asm.Advice;
+
+public class BoundedQueueBasedMessageQueueQueueAdvice {
```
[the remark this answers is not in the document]
worekleszczy: Names of advices follow pattern `{$ClassName}{$MethodName}Advice` - here `ClassName: BoundedQueueBasedMessageQueue` and `MethodName: Queue`

otel-extension/src/main/scala/io/scalac/mesmer/agent/util/dsl/Model.scala
worekleszczy: Removed :)

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/actor/Instruments.scala
```diff
@@ -0,0 +1,83 @@
+package io.scalac.mesmer.otelextension.instrumentations.akka.actor
+import io.opentelemetry.api.GlobalOpenTelemetry
+import io.opentelemetry.api.metrics.LongCounter
+import io.opentelemetry.api.metrics.LongHistogram
+import io.opentelemetry.api.metrics.MeterProvider
+
+trait Instruments {
```
lgajowy, line 7: So now, the `trait Instruments`, `object Instruments` and `InstrumentsProvider` is just `object Instruments` with some extra steps. We don't replace the instruments in testing as it turned out to be a no-go. 

Please replace with just an `object Instruments`. Please rename it to `AkkaActorInstruments` for more clarity

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/actor/Instruments.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  def failedMessages: LongCounter
+
+  def processingTime: LongHistogram
+
+  def unhandled: LongCounter
```
lgajowy, line 13: please rename to unhandledMessages

lgajowy: and apply to all message-related metrics

otel-extension/src/main/java/akka/actor/impl/typed/SupervisorHandleReceiveExceptionAdvice.java
lgajowy: bump

otel-extension/src/main/java/akka/actor/impl/typed/SupervisorHandleReceiveExceptionAdvice.java
```diff
@@ -0,0 +1,45 @@
...
+
+    if (Objects.nonNull(state) && Objects.nonNull(attributes)) {
+      state.setFailed();
+      instruments.failedMessages().add(1L, attributes);
+    } else {
+      // this should never happen
+      classicContext
+          .system()
+          .log()
+          .warning("Couldn't find mesmer actor cell state inside ActorCell");
```
lgajowy, line 42: This should be an "error" if we're not throwing exceptions (see above comment)

otel-extension/src/main/java/akka/actor/impl/ActorCellReceivedAdvice.java
```diff
@@ -0,0 +1,44 @@
...
+    Attributes attrs = VirtualField.find(ActorContext.class, Attributes.class).get(self);
+    Instruments instruments = InstrumentsProvider.instance();
+    ActorCellInstrumentationState state =
+        VirtualField.find(ActorContext.class, ActorCellInstrumentationState.class).get(self);
+
+    if (Objects.nonNull(attrs) && Objects.nonNull(state)) {
+      instruments.processingTime().record(interval, attrs);
+      /*
+         Here we check it there was an exception and if TypedInstrumentation already taken care of this
+      */
```
lgajowy, line 37: Theres
```suggestion

```

It's clear what the code is doing so the comment is redundant

otel-extension/src/main/java/akka/actor/impl/ActorCellReceivedAdvice.java
```diff
@@ -0,0 +1,44 @@
...
+      @Advice.Thrown Throwable exception,
+      @Advice.Enter long started) {
+    long interval = Interval$.MODULE$.toMillis(System.nanoTime() - started);
+
+    Attributes attrs = VirtualField.find(ActorContext.class, Attributes.class).get(self);
+    Instruments instruments = InstrumentsProvider.instance();
+    ActorCellInstrumentationState state =
+        VirtualField.find(ActorContext.class, ActorCellInstrumentationState.class).get(self);
+
+    if (Objects.nonNull(attrs) && Objects.nonNull(state)) {
```
lgajowy, line 33: This kind of check seems to be used quite often - could you create a function for it?

otel-extension/src/main/scala/io/scalac/mesmer/agent/util/dsl/package.scala
```diff
@@ -0,0 +1,72 @@
...
+import net.bytebuddy.description.method.MethodDescription
+import net.bytebuddy.matcher.ElementMatcher
+import net.bytebuddy.matcher.ElementMatchers
+
+package object dsl {
+
+  type MethodMatcher = ElementMatcher[MethodDescription]
+  type TypeMatcher   = ElementMatcher[TypeDescription]
+
+  // THIS IS A HACK BUT IT FORCES TYPE BOUND TO EVALUATE TO THE LOWEST POSSIBLE TYPE AND NOT NOTHING
```
lgajowy, line 17: No need for caps. Please fix

lgajowy:
```suggestion
  // Hack: it forces type bound to evaluate the lowest possible type and not "nothing".
```

otel-extension/src/main/scala/io/scalac/mesmer/agent/util/dsl/package.scala
```diff
@@ -0,0 +1,72 @@
...
+      typeDesc: ElementMatcher[_ >: TypeDescription]
+    ): ElementMatcher.Junction[T] = ElementMatchers.takesArgument[T](pos, typeDesc)
+
+  }
+
+  /**
+   * Factory method but with that will result in proper type inference - here we make more strict requirements for types
+   * of matchers but is should not limit any capabilities of [[ElementMatchers]]
+   */
+  def typeInstrumentation(`type`: TypeMatcher): TypeAppliedInstrumentation = new TypeAppliedInstrumentation(`type`)
```
lgajowy, line 57: I introduced the `Instrumentation` dsl in https://github.com/ScalaConsultants/mesmer/pull/425. Please use that one

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/akka/actor/AkkaActorAgent.scala
```diff
@@ -1,184 +1,130 @@
...
+  val actorUnhandledAdvice: TypeInstrumentation =
+    typeInstrumentation(
+      matchers.named("akka.actor.Actor")
+    )(
+      matchers.named("unhandled"),
+      "akka.actor.impl.ActorUnhandledAdvice"
+    )
 
-  private val mailboxInstrumentation = {
+  // mailbox
```
lgajowy, line 85:
```suggestion
```

lgajowy (changes requested):

otel-extension/src/it/scala/io/scalac/mesmer/agent/utils/OtelAgentTest.scala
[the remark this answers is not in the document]
lgajowy: see: https://github.com/ScalaConsultants/mesmer/pull/418/files#r923203400

otel-extension/src/main/java/akka/actor/impl/ActorCellReceivedAdvice.java
worekleszczy: I don't really see how this will make the code clearer - a couple of null checks state pretty well what is going on and and It's not not a lot of repetition

otel-extension/src/main/java/akka/actor/impl/ActorCellReceivedAdvice.java
worekleszczy: I disagree - I must say that this part took me a while to figure out from the old code that's why I left this comment and there is more in depth explanation of the process in `akka.actor.impl.typed.SupervisorHandleReceiveExceptionAdvice`  . We can change this to something else though like: 
```
       To keep failed messages metric consistent we must not increase the counter if type instrumentation already handled it
         Check akka.actor.impl.typed.SupervisorHandleReceiveExceptionAdvice for more details
```
WDYT @lgajowy

core/src/main/scala/io/scalac/mesmer/core/akka/model/AttributeNames.scala
```diff
@@ -0,0 +1,8 @@
+package io.scalac.mesmer.core.akka.model
+
+object AttributeNames {
```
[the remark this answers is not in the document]
worekleszczy: If we follow this rule rigorously we'll end up with a single module and this will get tricky to work with. I imagine that in the future we will get modules for all every technology that we support with core module shared by all of those - this is something that might be shared for all instrumentations. I don't really like the idea of putting everything in the single module TBH

core/src/main/scala/io/scalac/mesmer/core/actor/ActorRefConfiguration.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import akka.actor.ClassicActorSystemProvider
+import akka.actor.typed.ActorRef
+import akka.actor.{ ActorRef => ClassicActorRef }
+import io.opentelemetry.api.common.Attributes
+import io.opentelemetry.api.common.AttributesBuilder
+
+import io.scalac.mesmer.core.akka.model.AttributeNames
+
+trait ActorRefConfiguration {
```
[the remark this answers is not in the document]
worekleszczy: same as above

core/src/main/scala/io/scalac/mesmer/core/akka/model/AttributeNames.scala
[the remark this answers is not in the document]
lgajowy: Note we can have clean separation not only on the module level but also on the package level.

Right now: 
 - there's no need for modules other than trying to be future-proof and the future requiring such a split will most probably never come
 - we have code split between modules so we don't really see at the first glance how much code is required to actually instrument something.
 - there's code that gets forgotten and unused (eg. JarNames object). Since we look at core less often, there's a smaller chance it will be removed/maintained etc
 - we have an additional build.sbt code to read. sbt code is hard to follow so if possible it should be as simple as possible (that's minor but still valid)

> we'll end up with a single module and this will get tricky to work with. 

why would that be tricky given a clean separation between packages?

I think a sensible common ground would be just to keep core at the necessary minimum and only keep things like typeclasses ("Traverse", "Combine", "encode"), utils (time related, randomness related etc), whereas akka related/zio related stuff should land in the extension directly. wdyt? So I'd still suggest AtttributeNames migration.

otel-extension/src/main/java/akka/actor/impl/ActorCellReceivedAdvice.java
lgajowy: Ok - now i see the intent more clearly and like this version better. 

Could you then extract this to a private method and utilize comment hyperlinks? IDEs will support that (and avoid name staying old when renaming). Something like this: 

```
/**
   * To keep failed messages metric consistent we must not increase the counter if type
   * instrumentation already handled it.
   *
   * @see akka.actor.impl.typed.SupervisorHandleReceiveExceptionAdvice for more details
   */
  private static void incrementFailedMessages(
      Throwable exception,
      Attributes attributes,
      Instruments instruments,
      ActorCellInstrumentationState state) {
    if (Objects.nonNull(exception) && !state.getAndResetFailed()) {
      instruments.failedMessages().add(1L, attributes);
    }
  }
```

otel-extension/src/main/java/akka/actor/impl/ActorCellReceivedAdvice.java
lgajowy: It's not only about cleanliness - this check is repeated 6 times in multiple advice classes, so we could:
 - reuse it and never forgot about checking one of the parameters
 - have it documented in one place, even if the method name would be self-documenting (eg. checkPreconditions).

However, shouldn't we throw when situations like this occur? If for some reason we will have attrs or state null we will silently fail. Shouldn't Mesmer "die" in case something as terrible as this happens?

lgajowy (approved):

PR #435, opened 2022-07-08
four small changes

scalac-mtk:
1) the "runExampleWithOtelAgent" sbt command chooses the correct main
method automatically now, like the streaming example does already.

2) prevent the environment variable HOSTNAME from overriding the default
seed node hostname (127.0.0.1 aka localhost).  this prevented the example
app from joining the cluster.

3) quiet the compilation warnings over '.take' vs '.take()' in the
actor events monitor.

4) quiet the implicit conversion warning in the istrumentation module
with a tightly-scoped scala.language.implicitconversions import.


lgajowy (approved):
👍 Thanks!

PR #436, opened 2022-07-11
420 execution metrics zio

### LINKED ISSUE #420: Auto-Instrument ZIO 2.0 execution metrics
[ZIO Executors](https://zio.dev/api/zio/internal/executionmetrics) expose some relevant metrics, such as: 
- workers_count,
- concurrency,
- executor_capacity,
- task_queue_size, 
- enqueued_count,
- dequeued_count

By auto-instrumenting the Executor's constructor, we can get a reference to it. Such a reference can be then used in asynchronous instrument callbacks to periodically collect execution metrics and report them via openTelemetry API. 

Proposal Link with code snippets on how to do that (see Part 1): [here](https://docs.google.com/document/d/18pR8oGcXXPyZF-L6CKch_ABzLRgtYJtFn-IHwrPCFx8/edit?usp=sharing)

Configuration: 
 - we should be able to turn the whole instrumentation on/off (that's for free thanks to openTelemetry)
 - nothing else should be configured for now in order to keep this ticket small and easy to review. In the future we can extend the config

Things to be ironed out: 
 - the proposal distinguishes the executors by calling executor.toString() which is a naive way of doing this. Is there a better way? 
 - is there a way to have an extra attribute: isBlocking (true/false). It seems beneficial to be able to distinguish blocking/non-blocking executors using attributes.

lgajowy:
Closes #420  
___

## Summary:

 - adds executor metrics
 - adds tests for the metrics

### DIFF
--- a/otel-extension/src/it/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOExecutorMetricsTest.scala
+++ b/otel-extension/src/it/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOExecutorMetricsTest.scala
@@ -1,0 +1,88 @@
+package io.scalac.mesmer.otelextension.instrumentations.zio
+
+import io.opentelemetry.api.common.AttributeKey
+import io.opentelemetry.sdk.metrics.data.{ MetricData, MetricDataType }
+import io.scalac.mesmer.agent.utils.OtelAgentTest
+import org.scalatest.flatspec.AnyFlatSpecLike
+import org.scalatest.matchers.should.Matchers
+import zio._
+
+import java.util.concurrent.{ SynchronousQueue, ThreadPoolExecutor, TimeUnit }
+import scala.jdk.CollectionConverters._
+
+class ZIOExecutorMetricsTest extends OtelAgentTest with AnyFlatSpecLike with Matchers {
+
+  val testProgram: ZIO[Any, Nothing, Long] = (for {
+    _ <- Random.nextInt
+  } yield ())
+    .repeat(Schedule.forever)
+
+  "ZIOExecutorMetrics" should "produce worker_count metric" in {
+    Unsafe.unsafe { implicit u =>
+      Runtime.default.unsafe.runToFuture(testProgram)
+
+      assertGaugeIsCollected("mesmer_zio_executor_worker_count")
+      assertGaugeIsCollected("mesmer_zio_executor_concurrency")
+      assertGaugeIsCollected("mesmer_zio_executor_dequeued_count")
+      assertGaugeIsCollected("mesmer_zio_executor_enqueued_count")
+      assertGaugeIsCollected("mesmer_zio_executor_size")
+      assertGaugeIsCollected("mesmer_zio_executor_capacity")
+      assertGaugeIsCollected("mesmer_zio_executor_capacity")
+    }
+
+    def assertGaugeIsCollected(name: String): Unit = assertMetrics(metricName = name) { case _ => () }
+  }
+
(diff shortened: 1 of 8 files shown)


otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOExecutorMetrics.scala
```diff
@@ -0,0 +1,74 @@
...
+
+object ZIOExecutorMetrics {
+
+  private val meter: Meter = GlobalOpenTelemetry.getMeter("mesmer")
+
+  private val executionMetricName: String => String = (suffix: String) => s"mesmer_zio_executor_$suffix"
+
+  def registerExecutorMetrics(executor: Executor): Unit =
+    Unsafe.unsafe { implicit u =>
+      val executorAttributes = Attributes.builder().put("executor", executor.toString).build()
```
worekleszczy, line 22: I think that we should try to distinguish between default `ZScheduler` and `zio.internal.Blocking.blockingExecutor` that should be consistent across multiple run of the application - right now if a service is restarted that run with default `Executor` `exector` attribute will change

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOExecutorMetrics.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  private def registerAsyncLongGauge(name: String, unit: String, description: String)(
+    callback: Consumer[ObservableLongMeasurement]
+  ): ObservableLongGauge =
+    meter
+      .gaugeBuilder(name)
+      .ofLongs()
+      .setUnit(unit)
+      .setDescription(description)
+      .buildWithCallback(callback)
```
worekleszczy, line 73: To protect resources we should add a deregistration logic on executor shutdown

otel-extension/src/it/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOExecutorMetricsTest.scala
```diff
@@ -0,0 +1,111 @@
...
+
+import java.util.concurrent.{ SynchronousQueue, ThreadPoolExecutor, TimeUnit }
+import scala.jdk.CollectionConverters._
+
+class ZIOExecutorMetricsTest extends OtelAgentTest with AnyFlatSpecLike with Matchers {
+
+  val testProgram: ZIO[Any, Nothing, Long] = (for {
+    _ <- Random.nextInt
+  } yield ())
+    .repeat(Schedule.forever)
```
worekleszczy, line 18: Maybe just 5 sec will be ok?

lgajowy: Wouldn't it potentially be flaky?

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOExecutorMetrics.scala
lgajowy: Good point. Do you think config parameters will be a good idea?

lgajowy: But, the question is how do we distinguish the blocking/non-blocking executor in the advice?

worekleszczy (approved):

PR #443, opened 2022-07-14
421 zio counters instrumentation

### LINKED ISSUE #421: Attach openTelemetry metrics to zio.metric.Counters
Proposal Link with code snippets on how to do that (see Part 2): [here](https://docs.google.com/document/d/18pR8oGcXXPyZF-L6CKch_ABzLRgtYJtFn-IHwrPCFx8/edit?usp=sharing)

ZIO defines Counters internally (eg fibersStarted up/down counter) . In order to be able to collect data collected by the internal conters via OpenTelemetry, it seems to be enough to create an asynchronous version of an otel counter, whit a callback using the ZIO Counter. Thanks to that the data it collects will be propagated to mesmer extension. 

Added value: Counters defined by users, that use the ZIO metrics Api will be provided with otel instruments as well. 

Goals:
 - propagating counter data to otel

NON-goals: 
 - excluding particular metrics collection. This should be planned and done in a separate issue
 - adding a sophisticated configuration. That can be postponed.

lgajowy:
Closes #421 

## Summary:

- instruments any ZIO counters (custom ones or provided by ZIO runitme)
- tests the instrumentations


lgajowy:
@worekleszczy pls note it's based on #420 's branch so you might want to review #420 first

lgajowy (commented):
This solution is not 100% correct. 

Currently, every counter will be instrumented ok, but it will never propagate tags/metric labels ("attributes" in otel naming). This is because whenever users want to create a metric with labels in ZIO, they use the following function: 

```
  // in zio.metrics.Metric

  /**
   * Returns a new metric, which is identical in every way to this one, except
   * the specified tags have been added to the tags of this metric.
   */
  final def tagged(extraTag: MetricLabel, extraTags: MetricLabel*): Metric[Type, In, Out] =
    tagged(Set(extraTag) ++ extraTags.toSet)
```

We need to find a way to propagate the MetricLabels in order to fully support zio metrics api.

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/advice/ZIOCounterAdvice.scala
```diff
@@ -0,0 +1,12 @@
...
+
+import net.bytebuddy.asm.Advice
+import zio.metrics.Metric.Counter
+
+import io.scalac.mesmer.otelextension.instrumentations.zio.ZIOMetrics
+
+object ZIOCounterAdvice {
+  @Advice.OnMethodExit
+  def counter(@Advice.Argument(0) zioMetricName: String, @Advice.Return counter: Counter[_]): Unit =
+    ZIOMetrics.registerCounterAsyncMetric(zioMetricName, counter)
```
worekleszczy, line 11: Why do we use async metrics here? IMO it would be better to use sync ones and avoid issue with porting histograms. We should be able to do so overriding `zio.metrics.Metric.fromMetricKey` to update otel instrument as well as zio counter. I think this method should be also portable for other metrics types supported in zio

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOMetrics.scala
```diff
@@ -0,0 +1,22 @@
...
+import io.opentelemetry.api.GlobalOpenTelemetry
+import io.opentelemetry.api.metrics.Meter
+import zio.Unsafe
+import zio.metrics.Metric.Counter
+
+object ZIOMetrics {
+
+  private val meter: Meter = GlobalOpenTelemetry.getMeter("mesmer")
+
+  private val executionMetricName: String => String = (suffix: String) => s"mesmer_zio_forwarded_$suffix"
```
worekleszczy, line 12: I would either just prefix it with `mesmer` or leave the metric name as is

worekleszczy (changes requested):

worekleszczy (approved):

lgajowy:
Added tags support by instrumenting the tagged method and fromMetricKey. It is still not ideal (sometimes you get a warning about duplicate instruments)  but for now it's ok.

PR #464, opened 2022-08-12
Fix/separate extension

worekleszczy:
## Summary:

Make otelExtension jar slimmer (without any dependencies apart from scala classes) and add conditional apply of streams transformations. At this point we no longer require extension to be on the classpath for other instrumentation to work, althought `core` is still required

### DIFF
--- a/otel-extension/src/main/java/io/scalac/mesmer/instrumentation/http/impl/OverridingRawPathMatcher.java
+++ b/otel-extension/src/main/java/io/scalac/mesmer/instrumentation/http/impl/OverridingRawPathMatcher.java
@@ -1,17 +1,18 @@
 package io.scalac.mesmer.instrumentation.http.impl;
 
-import akka.http.scaladsl.server.Directive;
 import akka.http.scaladsl.server.PathMatcher;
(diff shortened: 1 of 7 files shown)


otel-extension/src/main/java/io/scalac/mesmer/instrumentation/http/impl/OverridingRawPathMatcher.java
```diff
@@ -1,17 +1,18 @@
...
 
 public class OverridingRawPathMatcher {
 
   @Advice.OnMethodExit
   public static void onExit(
-      @Advice.Argument(0) PathMatcher<?> matcher,
-      @Advice.Return(readOnly = false) Directive<?> result) {
+      @Advice.Argument(0) Object matcher,
+      @Advice.Return(readOnly = false, typing = Assigner.Typing.DYNAMIC) Object result) {
+    //
```
lgajowy, line 14:
```suggestion
```

lgajowy (approved):

lgajowy (changes requested):
One more fix is needed - the stream example does not work now due to the fact that the example subproject imports only core. I'm going to cover that now

lgajowy:
outdated. closing

PR #468, opened 2022-08-16
Fixed snapshot metric - it had a wrong name

lgajowy:
Closes #xxx 
___

## Summary:

Fixes the incorrect metric name


ronaldSC-2022 (approved):
LGTM

PR #494, opened 2022-08-31
483 reinstate actors terminated metric

### LINKED ISSUE #483: Reinstate actors-terminated metric.
This metric is missing after removing akka extension from akka actor metrics. To be reinstated but now without using the extension

lgajowy:
Closes #483 

## Summary:

- Introduces a "micro" extension that is automatically spawned for actor lifecycle monitoring (created/terminated metrics).
- the extension uses an "event stream" instead of a custom EventBus that makes the code more concise and still doing the same thing
- extra tests and utils
- refactoring


build.sbt
```diff
@@ -187,7 +187,7 @@ lazy val example = (project in file("example"))
     commands += runStreamExampleWithOtelAgent,
     commands += runZioExampleWithOtelAgent
   )
-  .dependsOn(extension)
+  .dependsOn(core, extension)
```
lgajowy, line 190: Extension is still needed so that we see the akka cluster metrics. I will work on a separate PR to decouple this - I didn't want to do too much now

jczuchnowski (approved):

otel-extension/src/main/java/akka/actor/impl/LocalActorRefProviderAdvice.java
```diff
@@ -1,12 +1,15 @@
...
 import net.bytebuddy.asm.Advice;
 
 public class LocalActorRefProviderAdvice {
 
   @Advice.OnMethodExit
-  public static void actorOfExit() {
-    InstrumentsProvider.instance().actorsCreated().add(1);
+  public static void actorOfExit(
+      @Advice.Argument(0) ActorSystem system, @Advice.Return ActorRef ref) {
+    system.eventStream().publish(new ActorLifecycleEvents.ActorCreated(ref));
```
Roiocam, line 13: Hi @lgajowy, I am curious what is the purpose on use EventStream instead of using Otel meter to measure it?

otel-extension/src/main/java/io/scalac/mesmer/otelextension/akka/MesmerAkkaActorInstrumentationModule.java
```diff
@@ -81,6 +82,15 @@ public List<String> getAdditionalHelperClassNames() {
         "io.scalac.mesmer.otelextension.instrumentations.akka.actor.util.EnvelopeContext$",
         "io.scalac.mesmer.otelextension.instrumentations.akka.actor.impl.otel.ActorCellInstrumentationState",
         "akka.actor.ProxiedQueue",
-        "akka.actor.BoundedQueueProxy");
+        "akka.actor.BoundedQueueProxy",
+        "io.scalac.mesmer.otelextension.instrumentations.akka.actor.extension.ActorLifecycleMonitorExtension",
```
Roiocam, line 86: The otel muzzle plugin could make this simple, but I am not sure it could apply to SBT.

otel-extension/src/main/java/akka/actor/impl/LocalActorRefProviderAdvice.java
lgajowy: We ultimately are using the otel meter. It's just that termination emits an akka Terminated() event so we need to intercept it somehow. We try to keep such event processing to the minimum now and rely on pure otel instruments but it's not always possible I think :)

otel-extension/src/main/java/io/scalac/mesmer/otelextension/akka/MesmerAkkaActorInstrumentationModule.java
lgajowy: Oh yeah! For sure. :) It's just temporary. I have plans to use the muzzle plugin but it's sadly not supported by sbt. I think we are going to end up migrating to gradle in some not-so-distant future as this is the tool that the otel opentelementry-java /opentelemetry-java-instrumentations projects use.

otel-extension/src/main/java/akka/actor/impl/LocalActorRefProviderAdvice.java
Roiocam: Yep. I think I got the point, rather than enhance another method like `ActorCell.stop()`, watching an Actor Lifecycle is the better option.

Thanks for your time in explaining.

PR #500, opened 2022-09-08
Update opentelemetry to 1.17.0

jczuchnowski:
Closes #xxx 
___

## Summary:

(please input the summary)


lgajowy:
The javaagent/agent-for-testing jars are missing now - that's likely causing problems with tests and example

PR #514, opened 2022-10-14
514 - bring back mailbox size metric

skipper1982:
Closes #514
___

## Summary:

(please input the summary)

### DIFF
--- a/otel-extension/src/main/java/akka/actor/impl/MailboxEnqueueAdvice.java
+++ b/otel-extension/src/main/java/akka/actor/impl/MailboxEnqueueAdvice.java
@@ -3,7 +3,7 @@ package akka.actor.impl;
 import akka.actor.ActorContext;
 import akka.dispatch.Mailbox;
 import io.opentelemetry.api.common.Attributes;
-import io.opentelemetry.instrumentation.api.field.VirtualField;
+import io.opentelemetry.instrumentation.api.util.VirtualField;
 import io.scalac.mesmer.otelextension.instrumentations.akka.actor.Instruments;
 import io.scalac.mesmer.otelextension.instrumentations.akka.actor.InstrumentsProvider;
 import net.bytebuddy.asm.Advice;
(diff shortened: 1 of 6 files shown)


otel-extension/src/main/java/akka/actor/impl/MailboxEnqueueAdvice.java
```diff
@@ -0,0 +1,27 @@
...
+import akka.dispatch.Mailbox;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.instrumentation.api.field.VirtualField;
+import io.scalac.mesmer.otelextension.instrumentations.akka.actor.Instruments;
+import io.scalac.mesmer.otelextension.instrumentations.akka.actor.InstrumentsProvider;
+import net.bytebuddy.asm.Advice;
+
+import java.util.Objects;
+
+public class MailboxEnqueueAdvice {
```
lgajowy, line 13: There's a method in `ActorCell`: 

```
 /**
   * If the actor isLocal, returns the number of "user messages" currently queued,
   * which may be a costly operation, 0 otherwise.
   */
  def numberOfMessages: Int
```

We could (maybe?) pass a reference to that field every time an actor is created. 

```
object LocalActorRefProviderAdvice {

  @OnMethodExit
  def actorOf(@Return ref: ActorRef, @Argument(0) system: ActorSystem): Unit =
    ??? // get the reference, the path and create the instrument for the actorRef (that contains the actorcell, which has the numberOfMessages method).

}
```

that advice could be attached to: 

```
private val localActorRefProviderInstrumentation: AgentInstrumentation =
    instrument("akka.actor.LocalActorRefProvider".fqcnWithTags("create"))
      .visit(LocalActorRefProviderAdvice, "actorOf")
```

(taken from v0.7.0)

everytime we create a new actor maybe we can setup an async instrument that would periodically poll for the value from `numberOfMessages`, thus taking the mailbox size. Gauge seems a good fit. Not sure about the performance of this (probably awful, but maybe not?).


@skipper1982 lmk what do you think. I realize this can be unclear so let's have a call if needed. :)

lgajowy (commented):
@skipper1982 I think I have an idea on how to use an async instrument for this metric (alternative approach). LMKWDYT. :)

lgajowy (approved):

lgajowy:
@adamgadomski I created an issue for a potential improvement of this metric: https://github.com/ScalaConsultants/mesmer/issues/615

For now merging as is.

PR #554, opened 2022-12-07
Bring back running-streams total and actors total metrics

### LINKED ISSUE #536: Bring back akka streams metrics

skipper1982:
Closes #536 
___

## Summary:

(please input the summary)


lgajowy (changes requested):
The solution seems to be working, but: 
 - it doesn't use the io.scalac.mesmer.agent.util.i13n.Instrumentation dsl but uses the old Agent API (not yet sure why - legacy reasons?)
 - has no tests for the otel extension, but I think nowadays we could easily add some basic ones (I mean OtelAgentTests)
 - are there any attributes that should be added (eg for actors running in streams, have a stream id attribute?) That is ofc if this is possible (not sure yet)
 
 
 I will try to cover these bullets now.

lgajowy (approved):

PR #576, opened 2023-01-23
Update/byte buddy 1.12.22

lgajowy:
Closes #xxx 
___

## Summary:

(please input the summary)


lgajowy:
This version of bytebuddy seems to be incompatible with currently used otel version. Will need to update it together.

Issue #580, opened 2023-01-26
Fix tests flakiness due to delta aggregation temporality


lgajowy:
There are some Ignored tests with a comment: `// TODO: Fix the test so that it works with DELTA aggregation`
relied on cumulative nature of instruments of InMemoryMetricExporter but now it was changed internally in the AgentForTesting to be DELTA: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5923/files/9ea0823de46a4ac6597a3cd588f78fa3af573bb0#r856619839

To be fixed or replaced with something better

UPDATE: In general, all integration tests that wait for a metric value can be flaky - revise them and fix.

lgajowy:
It's about 6 tests now

PR #605, opened 2023-02-15
Roll back agent-for-testing temporality change from delta to cumulative

ptrdom:
Closes https://github.com/ScalaConsultants/mesmer/issues/580
___

## Summary:

It would seem that cumulative temporality is best fit for async code testing style typical for Scala codebases which are also used in this codebase. This change adds a custom-built agent-for-testing.jar that contains only one change:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/6fe3299f52f8c90859a8beecb70bda3fbbf43e09/testing/agent-exporter/src/main/java/io/opentelemetry/javaagent/testing/exporter/OtlpInMemoryMetricExporter.java#L37
`AggregationTemporality.DELTA` -> `AggregationTemporality.CUMULATIVE`.


lgajowy (commented):
One test still fails. Besides the other tests, ignored temporarily due to constant failures due to the temporality were not un-ignored (see the issue description). What's the behavior when you un-ignore them?

lgajowy:
Other than that: going this way will require us to rebuild the agent on our own each time we bump the version. Temporarily, I'd be ok with that but a short readme mentioning that this is required may be useful, I guess. 🤔

lgajowy (approved):

PR #612, opened 2023-02-17
Run examples in CI

ptrdom:
## Summary:

Things done in this PR:
1. Each example split into its own module.
1. Added `StdIn.readLine()` interruptions to example programs.
1. Each example is ran in CI.
1. Run build and example CI steps on both Linux and Windows.


ptrdom:
image

I think the check is caused by https://github.com/orgs/community/discussions/26698 - job names because of per OS runs.

build.sbt
```diff
@@ -49,7 +49,7 @@ lazy val all: Project = (project in file("."))
     name           := "mesmer-all",
     publish / skip := true
   )
-  .aggregate(extension, otelExtension, example, core, testkit)
+  .aggregate(extension, otelExtension, core, testkit, exampleAkka, exampleAkkaStream, exampleZio)
```
lgajowy, line 52: Yes! Finally someone had a reason to split the examples to separate modules. Thanks! :)

build.sbt
```diff
@@ -150,11 +150,31 @@ lazy val otelExtension = (project in file("otel-extension"))
...
+  )
+)
+
+lazy val exampleAkka = (project in file("examples/akka"))
+  .settings(exampleCommonSettings)
   .settings(
-    name           := "mesmer-akka-example",
-    publish / skip := true,
+    name := "mesmer-akka-example",
     libraryDependencies ++= {
```
lgajowy, line 178: What would you say about cleaning this up and moving the dependencies to Dependencies.scala like the rest of the deps?

lgajowy: eg. have zioExampleDependencies / akkaStreamingExampleDependencies etc in Dependencies.scala and reduce duplication in that file?

build.sbt
```diff
@@ -178,24 +198,51 @@ lazy val example = (project in file("example"))
...
+  .settings(exampleCommonSettings)
+  .settings(
+    name := "mesmer-akka-stream-example",
+    libraryDependencies ++= {
+      akka ++
+      logback ++ Seq(
+        "io.opentelemetry"    % "opentelemetry-sdk-extension-autoconfigure" % OpentelemetryAlphaMinor0Version,
+        "io.grpc"             % "grpc-netty-shaded"                         % "1.53.0",
+        "org.wvlet.airframe" %% "airframe-log"                              % AirframeVersion
+      )
```
lgajowy, line 219: The repetition could be avoided (see above)

examples/zio/src/main/scala/example/SimpleZioExample.scala
```diff
@@ -0,0 +1,17 @@
...
+import scala.io.StdIn
+
+object SimpleZioExample extends ZIOAppDefault {
+
+  override def run: ZIO[Any, Throwable, Any] = ZioProgram
+    .findTheMeaningOfLife(3, Int.MinValue, Int.MaxValue)
+    .race(ZIO.attempt(StdIn.readLine()))
+    .provide(
+      Runtime.enableRuntimeMetrics, // NOTE: refactored by following this zio-metrics-connectors example ( https://github.com/zio/zio-metrics-connectors/blob/zio/series2.x/core/jvm/src/test/scala/zio/metrics/connectors/SampleApp.scala#L15-L71 )
+      DefaultJvmMetrics.live.unit // NOTE: DefaultJvmMetrics.live collects the same JVM metrics as the Prometheus Java client's default exporters
```
lgajowy, line 15: While we're at it, let's remove the comments:

```suggestion
      Runtime.enableRuntimeMetrics,
      DefaultJvmMetrics.live.unit
```

the link is dead anyway and zio devs should be able to know what is going on.

lgajowy (changes requested):
Some really minor, boy-scout things. Thanks!

ptrdom:
Replicated the issue with `run / fork := true`  in Windows machines on CI, posted in Scala Discord - https://discord.com/channels/632150470000902164/922600050989875282/1077243996159614996 - let's see if we get any replies.

build.sbt
ptrdom: I do not see what is really solved by having `zioExampleDependencies`, `akkaStreamingExampleDependencies`, so I decided not to go with them, but I think I removed the last of duplication in example project settings. I am open to changing my mind though. 

Maybe moving all the library definitions to the `Dependencies.scala` looks nice, but maybe for example project it does not matter as much, especially if those dependencies are only used for examples. I guess we just have to agree on something and stick to it 😄

ptrdom:
> Replicated the issue with `run / fork := true` in Windows machines on CI, posted in Scala Discord - https://discord.com/channels/632150470000902164/922600050989875282/1077243996159614996 - let's see if we get any replies.

Not sure if I will get a reply for this, so I just disabled forking in Windows.

build.sbt
```diff
@@ -172,30 +189,40 @@ lazy val example = (project in file("example"))
...
+    mainClass := Some("example.SimpleStreamExample"),
+    run / javaOptions ++= Seq(
+      s"-Dotel.service.name=mesmer-stream-example",
+      s"-Dotel.metric.export.interval=5000"
+    )
+  )
+  .dependsOn(core)
 
-      keys
-    },
```
ptrdom, line 195: This feeding of `System.getProperties` to `javaOptions` was breaking the Windows classpath. It does not seem necessary for examples to work.

build.sbt
lgajowy: Might be personal preference but I feel like it's nicer to have all the deps in a separate file that is dedicated to deps only. That being said, if you're not convinced I'm ok with leaving as is because it's not very important to me.

lgajowy (approved):

lgajowy:
Looking at the issue with build status now....

lgajowy:
I've setup all the jobs to be required except the mend security check

PR #614, opened 2023-02-22
Fix assembling. Move disabling from "all" to specific modules

lgajowy:
Closes #xxx 
___

## Summary:

Fixes sbt assembly command being called from root of the proejct

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -45,7 +45,6 @@ addCommandAlias("testAll", "test; IntegrationTest/test")
 val projectRootDir = all.base.absolutePath
 
 lazy val all: Project = (project in file("."))
-  .disablePlugins(sbtassembly.AssemblyPlugin)
   .settings(
     name           := "mesmer-all",
     publish / skip := true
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -174,6 +173,7 @@ def exampleCommonSettings = Seq(
 )
 
 lazy val exampleAkka = (project in file("examples/akka"))
+  .disablePlugins(sbtassembly.AssemblyPlugin)
```
lgajowy, line 176: NOw I get: 

```
 there are 3 keys that are not used by any other settings/tasks:
[warn]  
[warn] * exampleAkka / mainClass
[warn]   +- /Users/lg/projects/mesmer/build.sbt:198
[warn] * exampleAkkaStream / mainClass
[warn]   +- /Users/lg/projects/mesmer/build.sbt:212
[warn] * exampleZio / mainClass
[warn]   +- /Users/lg/projects/mesmer/build.sbt:226
```

@ptrdom any hints? Or is this just an sbt oddity?

ptrdom: No, this is an actual problem. But we do not need that setting - those modules only have on main class in them.

ptrdom (approved):
LGTM

PR #635, opened 2023-03-08
Document actor paths configuration

### LINKED ISSUE #487: Document configuration of actorPath attribute in Actor metrics
https://github.com/ScalaConsultants/mesmer/discussions/381#discussioncomment-2610130 shows how it is done. Note that there might have been some changes in the newest version so before documenting testing is needed.

mBut:
Closes #487 
___

## Summary:

This PR adds documentation for the akka actors metrics grouping and performs little clean up in the associated codebase.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,0 +1,17 @@
+---
+sidebar_position: 3
+---
+
+# Configuration
+
+There are multiple ways to configure the Agent, including system properties, environment variables, and providing the configuration file.
+
+To provide the configuration file to the agent you can either `-Dotel.javaagent.configuration-file=<path-to-file>` or `OTEL_JAVAAGENT_CONFIGURATION_FILE` environment varaible.
+
+## OTEL namespace options
+
+You can find the full list of available configuration options [here](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md)
+
+## Mesmer namespace options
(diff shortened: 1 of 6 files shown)


docs/configuration.md
```diff
@@ -0,0 +1,70 @@
...
+
+To provide the configuration file to the agent you can either `-Dotel.javaagent.configuration-file=<path-to-file>` or `OTEL_JAVAAGENT_CONFIGURATION_FILE` environment varaible.
+
+## OTEL namespace options
+
+You can find the full list of available configuration options [here](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md)
+
+## Mesmer options
+
+### Akka actors grouping
```
mBut, line 17: @lgajowy I described the actual behaviour of the current implementation as is, but the implementation it's what bothers me. For most paths definitions there is no distinction between `instance` and `group` option behaviour. Similarly different rules can lead to the same results and it's quite confusing which definition is preferred. 

I believed the current implementation is the result of incremental evolution of the requirements. I suggest, for some future version of Mesmer, to define and document the exact specification on path matching first and then rewrite the entire matching logic from scratch.

lgajowy: Oh, yeah. Couldn't frame that better. Thanks for that input!

lgajowy (approved):

Issue #634, opened 2023-03-08
Refactor ZIO OTEL metrics to use `ConcurrentMetricRegistry`


ptrdom:
Our current approach to instrumenting ZIO metrics breaks if their mapping functions are used. A better approach would be to interact directly with `ConcurrentMetricRegistry`. It does provide two approaches - pull based `snapshots` and push based `listeners`. Currently push based 'listeners' have at least one issue with them that makes them non-ideal to use - see https://github.com/zio/zio/issues/7897#issuecomment-1458206154. We should implement pull based `snapshots` approach now, maybe consider implementing push based `listeners` approach in the future and allow users of the Mesmer to toggle which one they want to use.

Should resolve https://github.com/ScalaConsultants/mesmer/issues/629 and https://github.com/ScalaConsultants/mesmer/issues/423.

ptrdom:
Seems like there is no implementation yet for synchronous gauge - https://github.com/open-telemetry/opentelemetry-specification/issues/2318.

PR #639, opened 2023-03-13
Refactor ZIO OTEL metrics to use `ConcurrentMetricRegistry`

ptrdom:
Closes https://github.com/ScalaConsultants/mesmer/issues/634


otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOMetrics.scala
```diff
@@ -1,45 +1,40 @@
...
 
-  def registerGaugeAsyncMetric(zioMetricName: String, counter: Gauge[_], attributes: Attributes): Unit =
+  def registerGaugeAsyncMetric(metricKey: MetricKey.Gauge): AutoCloseable =
     meter
-      .gaugeBuilder(metricName(zioMetricName))
-      .buildWithCallback(_.record(unsafeGetGaugeValue(counter), attributes))
+      .gaugeBuilder(metricName(metricKey.name))
+      .buildWithCallback(_.record(unsafeGetGaugeValue(metricKey), buildAttributes(metricKey.tags)))
 
   def buildAttributes(metricLabels: Set[MetricLabel]): Attributes = {
```
lgajowy, line 28:
```suggestion
 private def buildAttributes(metricLabels: Set[MetricLabel]): Attributes = {
```

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryPoller.scala
```diff
@@ -0,0 +1,60 @@
...
+import scala.concurrent.duration._
+import scala.util.control.NonFatal
+
+import io.scalac.mesmer.otelextension.instrumentations.zio.ZIOMetrics._
+
+class ConcurrentMetricRegistryPoller {
+
+  private val timer = new Timer()
+
+  private val instruments = mutable.HashMap.empty[MetricKey[MetricKeyType], AutoCloseable]
```
lgajowy, line 19: Looking at the VirtualFIeld docs now: 

```
/**
 * Represents a "virtual" field of type {@code F} that is added to type {@code T} in the runtime.
 *
 * <p>A virtual field has similar semantics to a weak-keys strong-values map: the value will be
 * garbage collected when their owner instance is collected. It is discouraged to use a virtual
 * field for keeping values that might reference their key, as it may cause memory leaks.
 *
 * @param <T> The type that will contain the new virtual field.
 * @param <F> The field type that'll be added to {@code T}.
 */
 ```
 
 It makes me wonder if we can use it for the instruments and have the resources cleaned by the VF itself (when MetricKey is collected, the instrument is as well?)

File: otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryPoller.scala [hunk 3]
```diff
@@ -0,0 +1,60 @@
...
+        }
+      } catch {
+        case NonFatal(_) =>
+        // TODO how to log the exception?
+      } finally
+        schedule()
+  }
+
+  def schedule(): Unit =
+    timer.schedule(task, 10.millis.toMillis) // TODO configurable polling interval?
```
lgajowy, line 57: Can you make it configurable in this PR?

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryPoller.scala
```diff
@@ -0,0 +1,60 @@
...
+  private def task = new TimerTask {
+    override def run(): Unit =
+      try {
+        val snapshot = ConcurrentMetricRegistryClient.snapshot()
+
+        snapshot.filter { metricPair =>
+          !instruments.contains(metricPair.metricKey)
+        }.foreach { metricPair =>
+          val autocloseable = metricPair.metricKey.keyType match {
+            case _: MetricKeyType.Counter =>
```
lgajowy, line 30: Is it possible to use the same OTEL async Instrument instance for ZIO metrics with the same name but different attr+value sets? There is a new API for that in OTEL (I think): 

```
    meter
      .counterBuilder(metricName(metricKey.name))
      .ofDoubles()
      .buildObserver()
 ```
 
 that allows for batch callbacks (callbacks for multiple measurements). 
 
 IMHO, if possible we should do it in a followup - just making sure if that makes sense

File: otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryPoller.scala [hunk 5]
```diff
@@ -0,0 +1,60 @@
...
+          autoCloseable.close()
+        }
+      } catch {
+        case NonFatal(_) =>
+        // TODO how to log the exception?
+      } finally
+        schedule()
+  }
+
+  def schedule(): Unit =
```
lgajowy, line 56:
```suggestion
  private def schedule(): Unit =
```

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryPoller.scala
```diff
@@ -0,0 +1,60 @@
...
+        snapshot.filter { metricPair =>
+          !instruments.contains(metricPair.metricKey)
+        }.foreach { metricPair =>
+          val autocloseable = metricPair.metricKey.keyType match {
+            case _: MetricKeyType.Counter =>
+              registerCounterAsyncMetric(metricPair.metricKey.asInstanceOf[MetricKey.Counter])
+            case _: MetricKeyType.Gauge =>
+              registerGaugeAsyncMetric(metricPair.metricKey.asInstanceOf[MetricKey.Gauge])
+            case _ =>
+              // TODO setup sync instruments for histograms
```
lgajowy, line 35: Please create an issue for that

lgajowy (changes requested):
Thanks!

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryPoller.scala
ptrdom: There is an `Autocloseable` that also needs to be closed, so garbage collection is not enough - we also need to execute a callback. 

I guess we could also fully tie this functionality to `MetricKey` lifecycle - setup instruments in their constructor, tear down on garbage collection or something like that. This would remove the need for polling. Though I do not think the polling is overly expensive, since it just processes an internal map. But maybe such approach would be somewhat cleaner?

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryPoller.scala
ptrdom: I guess so, but signature does look funky there:
- https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/metrics/Meter.java#L141-L147
- https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/metrics/ObservableDoubleMeasurement.java
The signatures do not make their use very clear for me.

Also, this would also conflict with a polling-free approach tied to lifecycle of `MetricKey` that I explained here - https://github.com/ScalaConsultants/mesmer/pull/639#discussion_r1135223341.

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryPoller.scala
ptrdom: https://github.com/ScalaConsultants/mesmer/issues/640

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryPoller.scala
lgajowy: > Autocloseable 

Oh right.

Let's stick with the current solution and then iterate on it if we find it annoying/wrong. For now it's more than enough

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryPoller.scala
lgajowy: Added to 0.9.0

PR #642, opened 2023-03-14
Add all classes from scala-reflect as helper classes for zio

lgajowy:
Closes #xxx 
___

## Summary:

adds all scala reflect classes as helper classes. Classes list was extracted from scala-reflect (find . -name "*.class") and prepared with some regex magic


lgajowy:
I think we should treat this as a temp solution until Muzzle is used

otel-extension/src/main/java/io/scalac/mesmer/otelextension/zio/ZIOHelpers.java
ptrdom, line 1: 👀 ❗

ptrdom (approved):
LGTM

Issue #640, opened 2023-03-14
Add sync instruments for histograms


ptrdom:
Once https://github.com/ScalaConsultants/mesmer/pull/639 is merged, the new ZIO Metrics code can be extended for histograms too.

PR #650, opened 2023-03-24
Add sync instruments for histograms

ptrdom:
Closes https://github.com/ScalaConsultants/mesmer/issues/640
___

## Summary:

Adds sync instrument for histograms. While working on this issue I had many issues with classloading, where tests would work, but running the ZIO example would have various issues, so I modified the `ConcurrentMetricRegistryAdvice` and related classes to be less sensitive to classloading eagerness and order.


ptrdom:
I also removed Scala Reflection, because I figured out how to use Java Reflection for our use cases, so we can drop this unnecessary dependency.

otel-extension/src/it/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOMetricsTest.scala
```diff
@@ -54,6 +58,10 @@ class ZIOMetricsTest
 
     assertCounterMetricValue("mesmer_zio_forwarded_zio_fiber_started", 2)
     assertCounterMetricValue("mesmer_zio_forwarded_zio_fiber_successes", 2)
+    assetHistogramData(
```
lgajowy, line 61: asset -> assert?

otel-extension/src/it/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOMetricsTest.scala
```diff
@@ -109,4 +148,9 @@ class ZIOMetricsTest
   private def getCounterAttributes(data: MetricData): Option[Attributes] =
     data.getDoubleSumData.getPoints.asScala.map(_.getAttributes).toList.headOption
 
+  private def assetHistogramData(metricName: String, block: HistogramData => Unit): Unit =
```
lgajowy, line 151: Not necessarily for now, but in general: the "assert..." methods could be extracted to some helper class

lgajowy (commented):
I need to look closer and play with it so I will finish reviewing this tomorrow. :)

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryClient.scala
```diff
@@ -32,33 +35,54 @@ object ConcurrentMetricRegistryClient {
...
 
-  private lazy val metricsRegistry = {
-    val moduleSymbol   = runtimeMirror.staticModule("zio.internal.metrics.package$")
-    val moduleMirror   = runtimeMirror.reflectModule(moduleSymbol)
-    val moduleInstance = moduleMirror.instance
-    val instanceMirror = runtimeMirror.reflect(moduleInstance)
-    val methodSymbol   = instanceMirror.symbol.info.member(ru.TermName("metricRegistry")).asTerm.getter.asMethod
-    val methodMirror   = instanceMirror.reflectMethod(methodSymbol)
-    methodMirror.apply()
+  private lazy val unsafe = {
```
lgajowy, line 81: Could you put the provate lazy vals above the class' functions?

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ConcurrentMetricRegistryClient.scala
```diff
@@ -32,33 +35,54 @@ object ConcurrentMetricRegistryClient {
...
-  private lazy val unsafe = {
-    val runtimeMirror  = ru.runtimeMirror(getClass.getClassLoader)
-    val moduleSymbol   = runtimeMirror.staticModule("zio.Unsafe$")
-    val moduleMirror   = runtimeMirror.reflectModule(moduleSymbol)
-    val moduleInstance = moduleMirror.instance
-    val instanceMirror = runtimeMirror.reflect(moduleInstance)
-    val methodSymbol   = instanceMirror.symbol.info.member(ru.TermName("unsafe")).asTerm.getter.asMethod
-    val methodMirror   = instanceMirror.reflectMethod(methodSymbol)
-    methodMirror.apply().asInstanceOf[zio.Unsafe]
+  def addListener(listener: MetricListener): Unit = {
```
lgajowy, line 38: So this way we: 
 - add our own custom zio metrics listener (a proxy instance)
 - everytime "updateHistogram" is called we can invoke our custom action (in our case update the otel sync histogram)
 - this can be extended with other metrics that we just no-op for now
 
 This is so nice!

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOMetrics.scala
```diff
@@ -25,16 +26,35 @@ object ZIOMetrics {
       .gaugeBuilder(metricName(metricKey.name))
       .buildWithCallback(_.record(unsafeGetGaugeValue(metricKey), buildAttributes(metricKey.tags)))
 
+  def registerHistogramSyncMetric(metricKey: MetricKey.Histogram): DoubleHistogram = {
+    val underlying = meter
+      .histogramBuilder(metricName(metricKey.name))
+      .build()
+    // workaround for classloading issues
```
lgajowy, line 33: What's the workaround here exactly? Can we fix the names to be more descriptive somehow?

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOMetrics.scala
```diff
@@ -25,16 +26,35 @@ object ZIOMetrics {
...
-    ConcurrentMetricRegistryClient.get(metricKey).get().count
+    client.get(metricKey).get().count
 
   private def unsafeGetGaugeValue(metricKey: MetricKey[MetricKeyType.Gauge]): Double =
-    ConcurrentMetricRegistryClient.get(metricKey).get().value
+    client.get(metricKey).get().value
+}
 
+object ZIOMetrics {
+  trait DoubleHistogram {
```
lgajowy, line 56: Unless I'm missing something, if you make `buildAttributes` implicit, there's no need for the custom trait and you can use the io.opentelemetry.metrics.DoubleHistogram object. Did you consider that? 

On the other hand implicit may be ugly/hard(er) to migrate to java in the future(?)

lgajowy (commented):
Looks great! 

One last thing: I was trying to find any histogram in our zio example that I've started on my machine. Is there any? If not, could you modify the zio example to have one (or two)?

ptrdom:
@lgajowy `zio_fiber_lifetimes` is a ZIO runtime metrics that should be in Prometheus. Though it seems that it is basically split by the collector to multiple different metrics, if I understand what is going on correctly.

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOMetrics.scala
ptrdom: For some reason returning `io.opentelemetry.api.metrics.DoubleHistogram` returned by `build()` causes advices to throw `ClassNotFound` erorrs. I observed `underlying` to be a common name that can be found for wrapped implementations in various Scala codebases, if you think it needs to be different then please specify. I will update the comment to describe the issue in more detail.

otel-extension/src/main/scala/io/scalac/mesmer/otelextension/instrumentations/zio/ZIOMetrics.scala
ptrdom: The need for this trait is explained in this comment - https://github.com/ScalaConsultants/mesmer/pull/650#discussion_r1150430960.

lgajowy (approved):

lgajowy:
🎉

PR #652, opened 2023-03-24
Setup scalafmt check for sbt project files

ptrdom:
`scalafmtCheckAll`  does not check sbt project files, `scalafmtSbtCheck` is needed for that.


lgajowy (approved):
Makes total sense!

PR #662, opened 2023-03-29
Add e2e tests for example projects

ptrdom:
This PR adds e2e tests that do the following actions for each example project:
- Start docker-compose located in `examples/docker/docker-compose.yaml`.
- `sbt run` the example project.
- Use Prometheus API to assert expectations on the metrics produced by example project.

Potential TODOs:
- [ ] e2e tests only check for existence of single metric out of a metric group - more metrics and assertions for their values could be added.
- [ ] e2e tests are located in `IntegrationTest` configuration of example projects, even though they do not need to be located under specific example project, because they do not depend on them during compile time - e2e tests could be moved to their own separate module to make examples more lightweight.


lgajowy:
>  e2e tests are located in IntegrationTest configuration of example projects, even though they do not need to be located under specific example project, because they do not depend on them during compile time - e2e tests could be moved to their own separate module to make examples more lightweight.

A separate module for e2e sounds good to me even now. Other than clean separation and hunting down regressions, I see some future development in that module (or maybe in a similar one?) to be considered: 
  - in this module, we could have baseline examples without the otel and collect the execution time
  - we could compare execution times of the same examples but with otel agent + extension to see how "heavy" mesmer is. 
  
Note that even if our metrics and adding OTEL do not obstruct the performance much, it's great to have numbers for the users so that they can see for themselves. Also, one of the reasons for many premature optimizations was the lack of such a tool, I think. 

I know it's a little offtopic but wdyt?

ptrdom:
@lgajowy I do have two observations in regards to these topics:

- Mixing example and test code introduces conflicting concerns. Ideally the examples should be bare minimum, simple applications, but tests often require additional code, harnesses and tricks to cover various scenarios. Trying to maintain these both concerns on same code can become a bit of a headache. But then again, ideally you do want your examples to be tested. I have not come up with the perfect solution for this in my OSS work, though I do have some ideas, mainly centered around tests  doing test code generation into example projects during test runtime.
- Performance comparisons require very tightly controlled execution environments, so unless we want to provision a dedicated machine with specific configuration to do these comparisons, the results and value we get is rather questionable. Unless we just have these performance tests for user's/developer's local workstation use only.

I will put these new e2e tests into their own new module.

ptrdom:
Recently added `run-examples` jobs need to be removed from `main` branch protection rules 😄

lgajowy:
Very valid concerns

> Mixing example and test code introduces conflicting concerns. 

Yeah - that's what I'd avoid too. The examples we have should stay as they are. I meant the tests in e2e to be used in different ways (performance tests/regression tests). Having said that I didn't check the tools for collecting performance data so maybe the whole "performance measuring effort" should be done differently with a dedicated tool.

> Performance comparisons require very tightly controlled execution environments

Oh yeah. I know that very well from the Beam project. :D For starters we should use our local machines and to generate the numbers for people we should indeed use something dedicated. Even on dedicated machines the results may vary slightly, but when you collect the historical results eg. daily, you have a general idea on how good/bad your performance is.

ptrdom:
Hmm, I wonder how complicated it would be to setup an on-demand performance test runner machine to automate collection of performance data. I know that on the cloud such machine would not be too cheap, even if we run it only once a day. Maybe we have a spare machine in the office? It could be setup as a GitHub Action runner to run the tests on schedule - for example I am thinking about setting up a RPi4 as a GitHub Action runner at my home for ARM builds and private projects 😄

lgajowy:
>  I know that on the cloud such machine would not be too cheap

We need exact numbers. Probably worth to investigate. Another mode would be to just test (pre) release to see if we have any regressions in comparison to previous release. For sure there are ways of optimizing this and finding a good middle ground. :) Heck, maybe even our private machines are ok? 

> Maybe we have a spare machine in the office? 

I'd be sad if we'd need to poke anyone in the office to turn it on/off, restart etc. 😅

e2e-test/src/it/scala/io/scalac/mesmer/e2e/ExampleAkkaStreamTest.scala
```diff
@@ -0,0 +1,32 @@
...
+    with ExampleTestHarness
+    with Matchers
+    with Eventually
+    with EitherValues {
+
+  "Akka Stream example" should {
+    "produce stream metrics" in withExample("exampleAkkaStream/run") { prometheusApi =>
+      eventually {
+        prometheusApi.assert(
+          "promexample_mesmer_akka_streams_running_streams",
```
lgajowy, line 21: I think you could add the tests for all the other metrics that we have to cover everything. it will also be a great documentation

lgajowy (changes requested):
Great stuff. Some small comments. 

Do you think that instead prometheus, we could use otel protocol and query that in the e2e tests? It would not involve setting up prometheus and testcontainers afaik making the setup easier to manage. I'm fine with current solution - just looking for areas to improve.

e2e-test/src/it/scala/io/scalac/mesmer/e2e/ExampleAkkaStreamTest.scala
lgajowy: With one exception - probably no sense to do that for zio as we just forward zio metrics. But for akka we have our own that maybe we should always check for existence

e2e-test/src/it/scala/io/scalac/mesmer/e2e/ExampleTestHarness.scala
```diff
@@ -0,0 +1,204 @@
...
+                Failure(new RuntimeException(s"sbt process exited with a non-zero exit code $exitValue"))
+              else Success(())
+            )
+          }
+        }
+      }
+
+      try {
+        try
+          Await.result(processHandlePromise.future, 120.seconds)
```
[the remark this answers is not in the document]
ptrdom: CI can be slow. 60 seconds was not enough for certain cases, 30 seconds is not even enough for Akka projects locally.

ptrdom:
I guess we could setup a mock localhost OTEL collector that would just accept and store whatever requests come in from the example application and then assert for certain data being in those requests. It is somewhat effort intensive to setup and won't speed up the tests that much - example application startup takes way longer than starting docker-compose. But that would be a more lightweight solution long term.

ptrdom:
> are we sure that it's not enough to use AgentTestRunner/LibraryTestRunner for such tests? 

That is a completely different setup then, I would argue is is not an E2E test then.

ptrdom:
Overall it is nasty that we **_have_** to package and run the example projects as in production to be sure that Mesmer works. If general Java/Scala development would require that, nobody would work with both of these languages 😅 Maybe Muzzle can solve that, but E2E tests like these are the best assurance until then.

build.sbt
```diff
@@ -128,37 +128,35 @@ lazy val otelExtension = (project in file("otel-extension"))
...
     name := "mesmer-akka-example",
     libraryDependencies ++= {
       akka ++
       scalatest.map(_ % "test") ++
       akkaTestkit.map(_ % "test") ++
-      akkaPersistance ++ Seq(
-        "io.circe"                      %% "circe-core"                        % CirceVersion,
-        "io.circe"                      %% "circe-generic"                     % CirceVersion,
-        "io.circe"                      %% "circe-parser"                      % CirceVersion,
+      akkaPersistance ++ circe ++ Seq(
```
lgajowy, line 159: persistance -> persistence

e2e-test/src/it/scala/io/scalac/mesmer/e2e/ExampleAkkaTest.scala
```diff
@@ -0,0 +1,26 @@
...
+  // Akka example does not produce following metrics:
+  // mesmer_akka_actor_failed_messages
+  // mesmer_akka_actor_unhandled_messages
+  // mesmer_akka_actor_dropped_messages
+  private val akkaMetrics = (Seq(
+    "mesmer_akka_actor_mailbox_size",
+    "mesmer_akka_actor_stashed_messages",
+    "mesmer_akka_actor_sent_messages",
+    "mesmer_akka_actor_actors_created",
+    "mesmer_akka_actor_actors_terminated"
```
lgajowy, line 16: I think you can add the persistence and cluster metrics here as well

ptrdom: I cannot see neither cluster nor persistence metrics produced by the Akka example. Is something missing in its setup?

build.sbt
ptrdom: This change is not a part of this PR, but sure, I can fix it on the way.

e2e-test/src/it/scala/io/scalac/mesmer/e2e/ExampleAkkaTest.scala
lgajowy: Hmm... it might be the case that you need to send a rest request to generate persistence metrics. Maybe we can rebuild the example so that the rest call is not needed? When it comes to cluster metrics - it's odd that they're not there.

PR #664, opened 2023-03-30
Cleanup event bus

### LINKED ISSUE #460: Remove EventBus from mesmer-akka-core
After discussion with @lgajowy, the purpose of this Discovery Ticket is to:

- Determine if we can remove `EventBus` (https://github.com/ScalaConsultants/mesmer/blob/main/core/src/main/scala/io/scalac/mesmer/core/event/EventBus.scala#L20-L32) from mesmer-akka-core.

- If the answer is **yes**, then I will create a "follow-up-kanban-ticket" for implementing this "EventBus removal from the mesmer-akka-core code".

lgajowy:
Closes #460 
___

## Summary:

- lots of refactoring. Mainly: moving the EventBus to the otelextension module from core
- bugfix: removed `PhasedFusingActorMaterializerAdvice`. It was generating ActorEvents that were only consumed in tests. That means that in the production use, we generated unconsumed traffic.

There's probably a lot more refactoring and core cleanup to be done but I decided to not do it all in one blow.


ptrdom (approved):
LGTM

PR #678, opened 2023-04-07
Instrument `http4s`

### LINKED ISSUE #667: Instrument `http4s`
Look into ways to instrument [http4s](https://github.com/http4s/http4s). Ideally both server and client should be instrumented.

Two possible ways of approaching the implementation

1) [MetricOps](https://github.com/http4s/http4s/blob/4728244a8e8fbc28644f1f93bd37de2263b3dc94/core/shared/src/main/scala/org/http4s/metrics/MetricsOps.scala):
    - Implement OTEL `MetricOps`.
    - Instrument advices that wrap `HttRoutes` with OTEL `MetricOps`.
2) Instrumenting `http4s` backends directly:
    - `ember` seems like the best candidate.

ptrdom:
Closes https://github.com/ScalaConsultants/mesmer/issues/667

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -76,6 +76,7 @@ lazy val otelExtension = (project in file("otel-extension"))
     excludeDependencies += "io.opentelemetry.javaagent" % "opentelemetry-javaagent-bootstrap",
     libraryDependencies ++= {
       zio.map(_ % "provided") ++
+      http4s.map(_ % "provided") ++
       openTelemetryExtension.map(_ % "provided") ++
       opentelemetryExtensionApi ++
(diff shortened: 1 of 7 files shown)


ptrdom:
TODO:
- [ ] Create an e2e test for `http4s`.
- [ ] Add tests for error cases. I have investigated doing that for each metric, but ran into an annoying issue where apparently `testRunner.clearAllExportedData()` does not actually clear exported metrics, so having multiple tests for single metric seemed impossible to do.

lgajowy:
TODO for future reference: 
 - an example project could be useful that shows the metrics

lgajowy (approved):

